PR #8, opened 2026-04-02
Ai assisted

yanchenko-igor:
Fix SKA coin-type data loss across the sync and API pipeline

This branch fixes a series of bugs where SKA amounts were silently dropped or zeroed out at various points in the block sync, database, and API layers. All changes are 
backend-only; no frontend modifications.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

Changes

fix: set coin_type on VinTxProperty in processTransactions
Every vin was stored with coin_type = 0 (VAR) even when spending an SKA output. The fix derives the vin's coin type from the transaction's outputs (single-coin 
assumption) and sets it on VinTxProperty. Includes a new extraction test asserting CoinType == 1 for a synthetic SKA-1 input.

fix: persist coin_amounts in blocks table so SKA data survives restart
Per-coin totals were computed at sync time and held in memory but never written to the DB. After a restart the cache was cold and SKA data disappeared from the UI. Adds 
a coin_amounts JSONB column to the blocks table, populates it via MsgBlockToDBBlock, and reads it back in all four retrieveBlockSummary* functions. Includes schema 
migration.

fix: display SKA amounts on tx and address pages
Six display bugs caused SKA amounts to show as 0 on transaction and address pages. Root cause: Value int64 is 0 for SKA outputs; the real amount lives in SKAValue string
which was never read. Fixes:
- VinTxProperty gains SKAValue string, populated in processTransactions
- ska_value TEXT added to the vins table (DDL, insert, select, migration)
- retrieveVinsByIDs now scans coin_type (was missing) and ska_value
- Explorer vout/vin loops use FormatSKAAmount for SKA outputs/inputs
- TxBasic.SKASent populated from dbTx0.SentByCoin
- ReduceAddressHistory skips VAR accumulators for SKA rows and sets AddressTx.SKAValue/CoinType instead

fix: expose coin_type and ska_value on API vout responses
GET /api/tx/{txid} returned "value": 0 for SKA outputs with no coin type indicator. Adds CoinType uint8 and SKAValue string to apitypes.Vout and apitypes.TxOut, 
populated from the chainjson.Vout RPC result in GetAPITransaction and GetAllTxOut.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

Testing
All new behaviour is covered by unit tests. go test ./... passes with zero failures (excluding tags requiring a live DB).


edshav (approved):
Code looks good. The SKA data pipeline fixes are correct and consistent throughout — vin coin type derivation, coin_amounts persistence, address history handling, and API exposure all follow the same pattern. Tests cover the key cases. Minor notes left in the review thread but nothing blocking. Approved.

PR #10, opened 2026-04-03
feat: persist per-coin tx count and size in blocks table

yanchenko-igor:
## Summary

Adds CoinTxStats — per-coin transaction count and total serialized size — to the blocks table, mirroring the existing coin_amounts pattern. After a restart with a cold 
cache, the homepage blocks table shows per-coin tx counts without requiring a re-sync.

## Changes

New type — dbtypes.CoinTxStats (tx_count int, size uint32) defined in db/dbtypes to avoid import cycles; aliased as apitypes.CoinTxStats.

db/dbtypes/conversion.go — Added blockCoinTxStats() and populated CoinTxStats in MsgBlockToDBBlock — the sync path that writes to the DB.

blockdata/blockdata.go — Added blockCoinTxStats() and populated CoinTxStats on BlockDataBasic and BlockExplorerExtraInfo in CollectBlockInfo — the API/cache path.

api/types/apitypes.go — Added CoinTxStats map[uint8]CoinTxStats to BlockDataBasic and BlockExplorerExtraInfo.

db/dbtypes/types.go — Added CoinTxStats map[uint8]CoinTxStats to Block.

db/dcrpg/internal/blockstmts.go — Added coin_tx_stats JSONB to CREATE TABLE, INSERT ($27), and all 6 SELECT statements that already included coin_amounts.

db/dcrpg/queries.go — Insert passes ToJSONB(dbBlock.CoinTxStats) as $27; all 4 retrieve functions scan and unmarshal coin_tx_stats.

db/dcrpg/upgrades.go — ALTER TABLE blocks ADD COLUMN IF NOT EXISTS coin_tx_stats JSONB.

db/dcrpg/pgblockchain.go — Added coinRowsFromSummary() that merges CoinAmounts + CoinTxStats into CoinRowData; updated both call sites (block detail page and websocket 
path) to use it instead of coinRowsFromAmounts().

## Bug fixes included

Two bugs were caught and fixed after initial implementation:

- MsgBlockToDBBlock was not computing CoinTxStats, so the column was always NULL in the DB — tx counts and sizes were always 0.
- One call site in the BlockInfo builder was still using coinRowsFromAmounts() instead of coinRowsFromSummary(), so TxCount and Size on CoinRowData were always 0 even 
when data was present.

## Tests

- TestBlockCoinTxStats_Mixed — block with one VAR tx and one SKA-1 tx; asserts correct counts and sizes per coin type.
- TestBlockCoinTxStats_Empty — empty block returns nil.
- schema_test.go — asserts coin_tx_stats JSONB is present in CreateBlockTable DDL.

## Migration

Existing databases are handled automatically by the upgrade path — no manual intervention needed.

### DIFF
--- a/.kiro/hooks/gh-pr-review.kiro.hook
+++ b/.kiro/hooks/gh-pr-review.kiro.hook
@@ -1,0 +1,13 @@
+{
+  "enabled": true,
+  "name": "GitHub PR Review",
+  "description": "When a GitHub PR URL is mentioned in a prompt, fetch the diff with gh CLI, write a review, and show it to the user for approval before posting",
+  "version": "1",
+  "when": {
+    "type": "promptSubmit"
+  },
+  "then": {
+    "type": "askAgent",
+    "prompt": "Check if the user's message contains a GitHub pull request URL (e.g. https://github.com/.../pull/N). If it does:\n1. Use `gh pr view N --json title,body,files,commits` and `gh pr diff N` to fetch the PR details and diff.\n2. Perform a thorough code review.\n3. Show the full review body to the user as a preview.\n4. Ask the user: \"Should I post this review?\" and wait for their explicit confirmation before running `gh pr review N --approve --body \"...\"`.\nDo NOT post the review without the user's explicit approval.\nIf no PR URL is present, do nothing."
+  }
+}
--- a/REWRITE_PLAN.md
+++ b/REWRITE_PLAN.md
@@ -1,7 +1,7 @@
 # Monetarium Explorer — Rewrite Plan
 
 ## Notes
-- **Every task is a separate commit.**
+- **Every task is a separate branch feature/name_of_the_feature.**
 - **Frontend tasks (7, 8): bare minimum for compatibility only — no polish.**
 
 ---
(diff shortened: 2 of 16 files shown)


edshav (approved):
Looks solid. A few notes:

**blockCoinTxStats duplication** — the function is implemented identically in both `db/dbtypes/conversion.go` and `blockdata/blockdata.go`. Intentional given the two separate paths (DB sync vs API/cache), but if the tx classification logic ever changes both copies need updating. Worth a shared helper long-term.

**Tx classification assumption** — both copies use "first SKA output wins" to assign a tx to a coin type. A tx with mixed VAR+SKA outputs gets counted as SKA only. The total tx count across coin types stays correct (each tx counted once), but the per-coin breakdown could be misleading. Worth a comment documenting the "one tx = one coin type" assumption.

**Silent unmarshal errors** — `_ = json.Unmarshal(coinTxStatsJSON, &bd.CoinTxStats)` is consistent with the existing `coinAmountsJSON` pattern. NULL from old rows before migration will just leave the field nil, which is the right behavior.

**TestBlockCoinTxStats_Mixed** — both txs are in the regular tree only. A stake-tree SKA tx would be a good edge case to add eventually.

Issue #18, opened 2026-04-06
Aggregator: Implement big.Int precision in ParseTxns


edshav:
Refactor ParseTxns to accumulate totalOut and totalSize per coin type using map[uint8]*big.Int. Prevent rounding errors by avoiding float64 conversion until the final broadcast step.

Part of #14

Issue #17, opened 2026-04-06
Collector: Implement SKA Detection in mempoolTxns


edshav:
Update the mempoolTxns loop to inspect msgTx.TxOut for v.SKAValue. Populate the new SKA fields if detected. Ensure zero-overhead for standard VAR transactions.

Part of #14

Issue #16, opened 2026-04-06
Core: Extend Mempool Models for SKA Awareness


edshav:
Modify MempoolTx in mempool/collector.go to include CoinType (uint8) and SKAAmount (*big.Int). Update MempoolInfo to include CoinStats map[uint8]CoinTxStats to carry per-coin aggregates.

Part of #14

PR #23, opened 2026-04-06
feat: per-coin mempool tracking with quota-aware fill status fixes #14 #16 #17 #18 #19

### LINKED ISSUE #14: Mempool Multi-Token Pipeline & High-Precision Engine
Implementation of the backend data structures and high-precision arithmetic required to track VAR and SKA tokens in the mempool. Includes the logic for block space allocation and the 4-color status engine.

yanchenko-igor:
## Summary

Mempool previously tracked only a single VAR-based TotalOut and a single LikelyMineable.Size. SKA transactions contributed 0 to all totals and fill bars were stubs. This
PR implements full per-coin tracking (tx count, size, amount) and quota-aware fill status, with several bugs fixed during implementation.

## Changes

txhelpers/txhelpers.go
- TotalOutFromMsgTx — guards on CoinTypeVAR; SKA outputs no longer corrupt the VAR total
- SKATotalsFromMsgTx — new function; returns per-SKA-coin output totals as atom strings for a single tx

explorer/types/explorertypes.go
- MempoolCoinStats — new struct (tx_count, size, amount)
- MempoolTx.SKATotals map[uint8]string — per-SKA amounts per tx; DeepCopy updated
- MempoolShort.CoinStats map[uint8]MempoolCoinStats — per-coin mempool totals; DeepCopy updated
- CoinFillData.Color replaced with Status string ("ok", "borrowing", "full")
- TrimmedMempoolInfo.CoinFills — added field and copied in Trim() so homepage template receives it

mempool/collector.go
- mempoolTxns — sets SKATotals on each MempoolTx via SKATotalsFromMsgTx; totalOut loop guards on VAR only
- ParseTxns — builds CoinStats after the main loop, grouping by coin type

mempool/monitor.go
- TxHandler — increments CoinStats for each new tx as it arrives; calls StoreMPData on all savers with a deep copy of inventory so fill bars update immediately without 
waiting for the next block
- mempoolTxns — sets SKATotals on each MempoolTx

mempool/mempoolcache.go
- DataCache.StoreMPData — guards against nil stakeData (passed when called from TxHandler)

cmd/dcrdata/internal/explorer/explorer.go
- fillColor removed; replaced by computeCoinFills with quota rules:
  - VAR: 10% of MaxBlockSize guaranteed; SKA types share 90% equally per active type
  - "ok" = within quota, "borrowing" = over quota but block not full, "full" = block full
- StoreMPData reads MaxBlockSize from BlockchainInfo (fallback 393216); guards nil pointer

cmd/dcrdata/views/home.tmpl — fill bar uses class="fill-{{.Status}}" instead of inline background:{{.Color}}

## Tests

- TestTotalOutFromMsgTx_Mixed, TestSKATotalsFromMsgTx, TestSKATotalsFromMsgTx_VAROnly
- TestParseTxns_CoinStats — one VAR tx + one SKA-1 tx → correct TxCount, Size, Amount per coin
- TestComputeCoinFills — 7 cases covering all status transitions and VAR/SKA independence
- TestCoinFillData — updated to use Status

### DIFF
--- a/REWRITE_PLAN.md
+++ b/REWRITE_PLAN.md
@@ -943,4 +943,80 @@ via the BlockInfo path has non-zero TxCount and Size when CoinTxStats is present
 Demo: Block detail page and websocket block events show correct per-coin tx_count and size in coin_rows. GET /api/block/{height} returns 
 "coin_tx_stats": {"0": {"tx_count": 5, "size": 2048}, "1": {"tx_count": 2, "size": 512}}.
 
+Task 19: Per-coin mempool tracking (size, amount, tx count)
+Commit: fix: per-coin tx count, size, and amount tracking in mempool
+
+Objective: Mempool currently tracks only a single VAR-based TotalOut float64 and a single LikelyMineable.Size int32. SKA transactions contribute 0 to all totals. Mirror 
+the block-level CoinTxStats pattern to track per-coin tx count, size, and amount in mempool.
+
+Root cause — three broken sites:
+
+1. txhelpers.TotalOutFromMsgTx sums v.Value (int64) for all outputs — SKA outputs have Value == 0, their amount is in v.SKAValue *big.Int, never read.
+2. MempoolTx has no per-coin amount field — only TotalOut float64 (VAR only).
+3. ParseTxns in collector.go accumulates out, _ := dcrutil.NewAmount(tx.TotalOut) into regularTotal, ticketTotal, etc. — all zero for SKA. LikelyMineable totals and 
+CoinFills in StoreMPData are therefore wrong for SKA.
+
+Changes:
+
+txhelpers/txhelpers.go — fix TotalOutFromMsgTx to guard on VAR only:
+go
+func TotalOutFromMsgTx(msgTx *wire.MsgTx) dcrutil.Amount {
+    var amtOut int64
+    for _, v := range msgTx.TxOut {
+        if v.CoinType == cointype.CoinTypeVAR {
+            amtOut += v.Value
+        }
+    }
+    return dcrutil.Amount(amtOut)
+}
+
+
+Add SKATotalsFromMsgTx(msgTx *wire.MsgTx) map[uint8]string — mirrors blockCoinAmounts but for a single tx, returns atom strings keyed by SKA coin type.
+
+explorer/types/explorertypes.go — add to MempoolTx:
+go
+SKATotals map[uint8]string `json:"ska_totals,omitempty"`
+
+Update DeepCopy to copy the map.
+
+Add MempoolCoinStats struct (mirrors CoinTxStats):
+go
+type MempoolCoinStats struct {
+    TxCount int   `json:"tx_count"`
+    Size    int32 `json:"size"`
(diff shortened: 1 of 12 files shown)


edshav (changes requested):
The implementation structurally meets the overall architectural goals of the epic (#14) and elegantly handles the space allocation logic (#19) by decoupling the presentation terminology (`fill-ok`, `fill-borrowing`, `fill-full`). However, I have identified a critical performance issue during aggregation and a logical bug in the realtime websocket events that we need to fix.

#### 1. Inefficient Memory Handling in `ParseTxns` (Issue #18 Violation)
In `mempool/collector.go` (around line 473), the loop iterates through the entire mempool and converts back and forth between string and `big.Int` sequentially for every SKA output inside the loop:

```go
a, _ := new(big.Int).SetString(s.Amount, 10)
b, _ := new(big.Int).SetString(amtStr, 10)
if a != nil && b != nil {
	s.Amount = new(big.Int).Add(a, b).String()
}
```
**Feedback:** Issue #18 specifically requested to accumulate totalOut and totalSize per coin type using a `map[uint8]*big.Int`. The current approach generates compounding string and memory allocations ($O(N)$ string manipulations in the hot path), creating unnecessary work for the garbage collector on large mempools.
**Required Change:** Please create a temporary `map[uint8]*big.Int` outside the loop, use native `Add()` directly internally, and only convert the aggregated totals into strings at the very end when generating `MempoolCoinStats`. 

#### 2. Stale `Amount` in Live Websocket Updates (`mempool/monitor.go`)
Inside `TxHandler`, the `p.inventory.CoinStats` map is properly updated with the incremental transaction's `TxCount` and `Size`, but the `Amount` field is entirely skipped:

```go
// Around line 611
if len(tx.SKATotals) == 0 {
    s := p.inventory.CoinStats[0]
    s.TxCount++
    s.Size += tx.Size
    p.inventory.CoinStats[0] = s // BUG: s.Amount is untouched!
} 
```
**Feedback:** Because `s.Amount` isn't updated incrementally here, live dashboard viewers will receive websocket payloads with stale/incorrect transaction amounts. They won't see correct values until a new block arrives and triggers `ParseTxns` to rebuild everything.
**Required Change:** Add logic to incrementally add the incoming transaction amount string into `s.Amount`. A small helper function that safely loads the string into a `big.Int` and sets the sum back might be useful here.

#### Minor Notes:
* **Issue #16 Clarification:** The original requirement suggested explicitly adding `CoinType` and `SKAAmount` back into `MempoolTx`. The PR instead uses `SKATotals map[uint8]string`. Since a transaction can technically bundle different SKA types, using a map is indeed the better design architectural pattern.
* **Nil `StakeData` Behavior:** In `TxHandler`, `StoreMPData` gets called with `stakeData = nil`. This intentionally bypasses `mempoolcache.go` updates while still triggering `explorerUI.go` updates. This is correct behavior, but just drop a `//` comment explaining why `DataCache` safely ignores it.

edshav (approved):
LGTM! The string-to-number-to-string compounding loop overhead is fixed, and real-time live events correctly handle Amount increments via the helper method. Thanks for the quick update.

Issue #30, opened 2026-04-07
Frontend: rewrite Mining section template and styles


edshav:
Update the home page template and SCSS for the Mining section.

What to change:
- Rename "POW Reward" subsection to "PoW VAR Reward". Change the green difficulty indicator label to "Next Block VAR Reward Reduction".
- Add "PoW SKA Reward" subsection. Render one line per SKA type via template range over `PoWSKARewards`, formatted as `00.000000000000000000 SKA-n per last block` (18 decimal places).

Done when: both subsections render correctly with mock data on light/dark themes at mobile and desktop widths.

Part of #25

PR #69, opened 2026-04-11
feat(mining): add empty state template for SKA rewards display

edshav:
- Add empty state template to display when no SKA rewards are available
- Update mining controller to render empty template when rewards array is empty
- Display user-friendly message "No SKA rewards available" in empty state
- Improve UX by providing visual feedback instead of blank container

Refs #30


edshav:
duplicated commit

PR #70, opened 2026-04-11
Feature/supply frontend

### LINKED ISSUE #32: Frontend: replace Distribution section with Supply section in template and styles
Update the home page template and SCSS to replace the Distribution section with Supply.

What to change:
- Remove the old Distribution section markup entirely.
- Add "VAR Coin Supply" subsection showing circulating VAR with the `(of ~47M)` target label.
- Add "SKA Coin Supply" subsection rendered via template range over `SKACoinSupply`. Each SKA block shows: `In Circulation`, `Total Issued`, `Total Burned` — values formatted to 18 decimal places.
- Section title is "Supply".

Done when: the new section renders correctly with mock data on light/dark themes at mobile and desktop widths, and no Distribution markup remains.

Part of #26

edshav:
feat(mining): add empty state template for SKA rewards display
- Add empty state template to display when no SKA rewards are available
- Update mining controller to render empty template when rewards array is empty
- Display user-friendly message "No SKA rewards available" in empty state
- Improve UX by providing visual feedback instead of blank container

feat(supply): rename distribution controller and update template bindings
- Rename distribution_controller.js to supply_controller.js
- Update data-controller binding from "distribution" to "supply" in home.tmpl
- Update data-action binding from "distribution#handleBlock" to "supply#handleBlock"
- Update data-supply-target attributes replacing data-distribution-target
- Add detailed specification document for supply section implementation
- Prepares frontend for backend supply data structure integration

feat(supply): extract supply section to dedicated template and controller
- Extract Supply section from `home.tmpl` into new `home_supply.tmpl` template
- Add `home_supply` to `commonTemplates` in `explorer.go` and `home_template_test.go`
- Rewrite Supply section markup with VAR and SKA subsections following `home_mining.tmpl` pattern
- Update `supply_controller.js` to remove dead Decred-inherited targets (`coinSupply`, `mixedPct`, `devFund`, etc.)
- Keep only `varCirculating` and `exchangeRate` targets for live updates on new blocks
- Update spec document to mark implementation complete
- Follows established pattern for modular home page sections


edshav:
the WebSocket newblock event payload (blockData.extra) only carries lightweight per-block data — things that change with every block like the current exchange rate and VAR circulating supply.

SKA supply figures (InCirculation, TotalIssued, TotalBurned) are aggregate database values that don't arrive in the block event payload. Updating them live would require either a separate API call on each block or adding them to the WebSocket message — neither of which is in scope for this issue. So they're server-rendered on page load and stay static until the next full page load.

If live SKA supply updates are wanted in the future, the block event payload would need to be extended on the backend to include ska_coin_supply, and the controller would need to re-render the SKA rows from that data.

cmd/dcrdata/public/images/favicon/site.webmanifest
```diff
@@ -1,15 +1,15 @@
 {
-  "name": "Monetarium Block Explorer",
-  "short_name": "Monetarium",
+  "name": "MyWebSite",
+  "short_name": "MySite",
```
yanchenko-igor: Let's keep Monetarium branding.

yanchenko-igor (changes requested):

cmd/dcrdata/public/images/favicon/site.webmanifest
edshav: fixed

yanchenko-igor (approved):

PR #72, opened 2026-04-13
add Docker support and update daemon config directory

yanchenko-igor:


edshav (approved):
### Review Summary

The PR introduces a robust Docker setup using multi-stage builds and a non-root user, which follows security and efficiency best practices. It also correctly aligns the exposed container port with the application's default (`7777`) and updates the default daemon home directory search path from `dcrd` to `monetarium`.

#### Key Observations:
- **Consistent Rebranding**: The shift from `dcrd` to `monetarium` for the app data directory is good, but there's a discrepancy with the existing `README.md` which uses `monetarium-node`.
- **Docker Security**: Adding the `explorer` non-root user is a great security improvement.
- **Port Alignment**: Updating the exposed port to `7777` correctly reflects the default configuration.

---

### Feedback

#### Dockerfile
- **Non-root user**: Excellent addition of a non-root user. This is a best practice for container security.
- **Port 7777**: Corrected from `9508`. Matches `defaultMainnetPort` in `config.go`.

#### README.md
- **Volume Mapping**: There is an inconsistency. The "Local Testing" section (Line 147) instructs users to create `~/.monetarium-node`. If a user follows those instructions and then tries to run the Docker container with the suggested command (`-v ~/.monetarium:...`), it won't find their config/certs. We should standardize on either `~/.monetarium` or `~/.monetarium-node` throughout the entire document and code.

#### cmd/dcrdata/config.go
- **dcrdHomeDir**: As noted in the README, if the daemon is named `monetarium-node`, using `monetarium-node` here would be more consistent with the existing documentation and likely the default behavior of the node itself. If we intend to change the standard to just `monetarium`, we should update all references in the README.

PR #74, opened 2026-04-14
feat: show Vote SKA Reward from the latest block containing SKA transactions

yanchenko-igor:
Summary: Persist Vote SKA Rewards across SKA-free blocks
This PR fixes an issue where the "Vote SKA Reward" section on the home page would become empty whenever the latest block contained no SKA transactions.
🛠 Changes
- Latest SKA Block Search: Instead of relying on the current tip block for PerBlock reward calculations, the backend now searches backwards through the recent block summary (last 30 days) to find the most recent block that actually distributed SKA rewards.
- Accurate Voter Calculation: To ensure the reward ratio is correct, the logic now retrieves the Voters count specifically from the identified "Latest SKA Block" rather than using the voters count from the current tip.
- Stability: This ensures that the "Vote SKA Reward" values remain visible and reflect the most recent actual reward event, providing a more consistent user experience.
🧪 Verification
- Verified that blocks without SKA rewards no longer clear the section.
- Verified that the displayed PerBlock reward is correctly calculated based on the voters in the block where the rewards were actually distributed.

### DIFF
--- a/cmd/dcrdata/internal/explorer/explorer.go
+++ b/cmd/dcrdata/internal/explorer/explorer.go
@@ -631,18 +631,31 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
 	}
 
 	coinTypes := txhelpers.SSFeeCoinTypes(toSummaries(sumYear))
+	for ct := range blockData.ExtraInfo.SSFeeTotalsByCoin {
+		coinTypes[ct] = struct{}{}
+	}
 
 	// Find the latest block with SKA fee data for PerBlock calculation.
 	var latestSkaBlock *apitypes.BlockDataBasic
 	var latestSkaVoters int64
-	for i := len(sum30) - 1; i >= 0; i-- {
-		if len(sum30[i].SSFeeTotalsByCoin) > 0 {
-			latestSkaBlock = sum30[i]
-			// Get full block info to retrieve the voters count for this specific block
-			if bInfo := exp.dataSource.GetExplorerBlock(ctx, latestSkaBlock.Hash); bInfo != nil {
-				latestSkaVoters = int64(bInfo.BlockBasic.Voters)
+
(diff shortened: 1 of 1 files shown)


edshav (commented):
This review was performed by **Gemini 3 Flash**.

The PR successfully implements the fallback to historical SKA rewards when the current block contains no SKA transactions. However, there are a few refinements needed to prevent regressions and ensure data accuracy:

1. **New Coin Types**: The removal of the loop (around line 635) that adds coin types from the current block is a regression. If a new SKA coin type appears for the first time in the current block, it will be omitted from the UI until a subsequent block is processed.
2. **Current Block Priority**: The implementation should check the current block (`blockData`) first. If it contains SKA data, that should be used as the "latest" instead of searching history. This ensures the UI always shows the most recent data and avoids an unnecessary DB query when current data is available.
3. **Efficiency**: Refactoring the fallback search to a "Check Current, then Fallback to History" pattern would be cleaner and more robust.

**Line-by-line Comments:**

* **File:** `cmd/dcrdata/internal/explorer/explorer.go`
* **Line 635:** This removes the logic that captures new coin types appearing in the current block. If a new SKA coin type is first used in this block, it will not be displayed. We should keep the loop to ensure the UI is always aware of all active coin types.
* **Line 639:** We should check the current block (`blockData`) first. If `blockData.ExtraInfo.SSFeeTotalsByCoin` is not empty, we can use it immediately as the "latest" data source. This avoids a redundant DB query (`GetExplorerBlock`) and ensures we don't display "latest" data from a previous block when current data is available.

edshav (approved):
This review was performed by **Gemini 3 Flash**.

The latest commit correctly addresses all previous feedback:
1. **Regression Fixed**: New coin types appearing in the current block are now correctly captured.
2. **Current Block Priority**: The logic now prefers current block data, ensuring the display is always up-to-date and avoiding unnecessary DB queries.
3. **Robust Fallback**: The historical search remains as a reliable fallback for SKA-free blocks.

The code is clean, efficient, and meets all requirements. Approving.

PR #122, opened 2026-04-21
feat: transaction detail page with multi-coin support

### LINKED ISSUE #118: Extend transaction API to support SKA transaction data (single-coin model)
## Problem Description
Currently, the transaction API (`/api/tx/decoded/{transaction_hash}`) and the core transaction data struct passed into `cmd/dcrdata/views/tx.tmpl` are tightly coupled to the legacy VAR (DCR) token model. 

This architectural assumption breaks for the introduction of SKA transactions for a few key reasons:
* **Missing Coin Context:** The current transaction data struct does not identify the token type it governs.
* **Implicit Precision Loss:** Critical monetary fields such as `.Total`, `.Fee.ToCoin`, `.AmountIn` (for inputs), and `.Amount` (for outputs) rely on native floating-point representations (`float64`). SKA tokens utilize 15 integer digits and 18 decimal digits. A standard floating-point number tops out around 15-17 total significant digits, causing immediate and critical data loss for SKA transfers.
* **Hardcoded Display Assumptions:** Relying on implicit properties forces templates to construct hardcoded "DCR" labels and heavily rely on helpers like `float64AsDecimalParts`, breaking rendering if the transaction describes an SKA transfer.

## Required Backend Changes
To respect the single-coin nature of transactions, we must evolve the underlying backend struct models without overcomplicating them with multi-asset arrays.

* **Add an explicit `CoinType` field:** Ensure the root transaction struct defines exactly which token dictates the context of the entire transaction payload. 
* **Deprecate `float64` amounts:** Remove float usage for `.Total`, `.Fee`, and all nested `Vin.AmountIn` and `Vout.Amount` sizes.
* **Introduce High-Precision Numeric Properties:** Replace the aforementioned amounts with a scalable data type. This should utilize either strings (e.g. `"10000.123456789012345678"`) or large integer representations (e.g. `*big.Int` representing atoms) to guarantee zero precision loss across service boundaries.
* **Consistent State Validations:** Introduce API-level data assertions to confidently guarantee that incoming parsed blocks/mempool transactions do not attempt to map mismatched token inputs and outputs into a single transaction structure. 

*(Note: We must explicitly avoid introducing multi-token aggregation maps like `map[coinType]amount` since each transaction handles one exact coin type.)* 

## API Contract Changes
For `TxShort`, `TrimmedTx`, and paths like `/api/tx/decoded/{hash}`:
* Include an explicit token identifier parameter (e.g., `"coin_type": 1`). 
* Update JSON serialization responses to represent all values as high-precision formats. It is highly recommended to encode extremely large atomic sums as strings across JSON boundaries to circumvent the limits of standard `Number` processing in client interpreters.
* Inputs (`vin`) and outputs (`vout`) must mirror the root payload’s context predictably, rather than individually identifying varying token flags.

## Template Impact
The data supplied to the `tx.tmpl` structure must evolve to prevent VAR assumptions:
* The transaction root data injected into the template requires access to `{{.Data.CoinType}}` or a matching property to conditionally replace hardcoded string labels (e.g., removing hardcoded `"DCR"` tags from headers and table columns).
* The struct payloads exposed to template helpers (currently fed natively via variables like `float64AsDecimalParts .Total`) must be replaced to provide precise un-truncated data formats, likely requiring strings to be piped into a designated high-precision template formatting module.

## Acceptance Criteria
- [x] The root transaction struct and API payload explicitly expose a `CoinType`.
- [x] All financial quantities (Totals, Fees, UI Inputs/Outputs) calculate and serialize without any `float64` dependency to manage 18-decimal tokens accurately.
- [x] Strict type validation guarantees a single transaction’s `Vin` and `Vout` never mix VAR and SKA representations.
- [x] API JSON payloads safely relay massive atomic numbers without implicit truncation by representing them as string literals.
- [x] The core backend model fed into templates enables dynamic rendering for either VAR or SKA solely by reading the payload's single-coin context constraint.

### LINKED ISSUE #119: Update `tx.tmpl` to render high-precision single-coin SKA transactions
### Goal
Update the transaction template (`tx.tmpl`) to support high-precision rendering for single-coin SKA transactions, leveraging the backend refactor in Issue 118.

### Updated Requirements (Matched to Issue 118)
- [x] Determine global transaction context using `.CoinType` from the transaction data.
- [x] Render "Total Sent" and "Fee" using high-precision `TotalRaw` and `FeeRaw` string fields for SKA tokens, bypassing legacy `float64` fields (`Total`, `Fee.ToCoin`).
- [x] Render Input (`Vin`) and Output (`Vout`) table rows using `ValueRaw` strings and the `skaDecimalParts` helper to maintain precise decimal alignment.
- [x] Ensure all hardcoded "DCR" strings are replaced with dynamic `{{coinSymbol .CoinType}}` labels.
- [x] Maintain legacy `float64` rendering paths only for VAR transactions (Type 0) to preserve backward compatibility.

### Status
This implementation is complete as part of the work for Issue 118, utilizing the direct high-precision string fields provided by the enhanced backend structs.

edshav:
## Summary

Implements full multi-coin (VAR/SKA) support on the confirmed transaction detail page (`/tx/{txid}`), replacing hardcoded `float64`/DCR rendering with high-precision atom strings and dynamic coin-aware display.

### API types (`api/types/apitypes.go`)
- Replaced the `Vin` alias for `chainjson.Vin` with a first-class `Vin` struct that includes `CoinType`, `ValueInRaw` (atom string), and a local `ScriptSig` type.
- Added `CoinType`, `Total`, `Fee`, `FeeRate`, and `FeeRateRaw` fields to `TxShort` and `TrimmedTx`. `Total` and `Fee` are atom strings to avoid float64 precision loss.
- Added `ValueRaw` (atom string) to `Vout`. The legacy `Value float64` field is retained but marked deprecated.

### Explorer types (`explorer/types/explorertypes.go`)
- Added `TotalRaw`, `FeeRaw`, `FeeRateRaw`, `Size`, and `CoinType` to `TxBasic`.
- Embedded `*VoteInfo` directly (was a named field).
- Added `ValueRaw` to `Vin` and `Vout`. Legacy `float64` fields marked deprecated.

### DB / blockchain layer (`db/dcrpg/pgblockchain.go`)
- `GetAPITransaction` now determines the transaction's `CoinType` from the first output.
- For VAR transactions: computes `Total` and `Fee` using existing `TotalVout`/`TxFeeRate` helpers, formatted as atom strings.
- For SKA transactions: computes `Total` from `SKATotalsFromMsgTx`, and `Fee`/`FeeRateRaw` by resolving each input's previous outpoint via the new `OutPointAddressesAll` helper and performing `big.Int` arithmetic.
- Vin mapping is now explicit (was a `copy()`) to populate `ValueInRaw`, `CoinType`, and `ScriptSig` per input.
- Vout mapping populates `ValueRaw` for both VAR (atom conversion) and SKA (string passthrough).
- `GetTrimmedTransaction` propagates the new `CoinType`, `Total`, `Fee`, `FeeRate`, and `FeeRateRaw` fields.

### Tx helpers (`txhelpers/txhelpers.go`)
- Added `OutPointInfo` struct carrying `Addresses`, `CoinType`, `Value` (VAR atoms), and `ValueSKA` (`*big.Int`).
- Added `OutPointAddressesAll` which returns the full `OutPointInfo` for a previous outpoint.
- Refactored `OutPointAddresses` to delegate to `OutPointAddressesAll` (no behaviour change for callers).

### Template function (`cmd/dcrdata/internal/explorer/templates.go`)
- Added `coinSymbol` template function that maps a `uint8` coin type to its ticker string via `cointype.CoinType.String()`.

### Transaction template (`cmd/dcrdata/views/tx.tmpl`)
- Layout: collapsed the two-column card layout to a single full-width column.
- Total Sent, Fee, and Fee Rate cells now branch on `CoinType`: VAR uses the existing `float64AsDecimalParts` path; SKA uses `skaDecimalParts` with the new atom-string fields.
- Mix count display uses `coinSymbol` instead of hardcoded "DCR".
- Vin/Vout table headers use `coinSymbol` instead of hardcoded "DCR".
- Vin amount cell branches on `CoinType` to render `ValueRaw` for SKA inputs.
- Vout amount and spent-status cells branch on `CoinType` to render `ValueRaw` for SKA outputs.

### Wiki (`wiki/transaction/`)
- Updated `flow.compact.md`, `flow.full.md`, `impact.md`, and `patterns.md` to reflect the "single-coin per transaction" invariant and replace "token/multi-token" language with "coin/multi-coin".
- Added the Core Invariant section to `flow.full.md`.

## Testing

- Existing unit tests continue to pass (no test files modified).
- Manual verification: VAR transactions render identically to before; SKA transactions now display correct high-precision amounts and the correct coin ticker.

### DIFF
--- a/api/types/apitypes.go
+++ b/api/types/apitypes.go
@@ -105,6 +105,7 @@ type Vin struct {
 	ScriptSig    *ScriptSig `json:"scriptSig,omitempty"`
 	Sequence     uint32     `json:"sequence"`
 	AmountIn     float64    `json:"amountin,omitempty"`
+	SKAAmountIn  string     `json:"ska_amountin,omitempty"`
 	BlockHeight  uint32     `json:"blockheight"`
 	BlockIndex   uint32     `json:"blockindex"`
 	CoinType     uint8      `json:"coin_type,omitempty"`
@@ -444,6 +445,7 @@ type VinShort struct {
 	BlockHeight   *uint32 `json:"blockheight,omitempty"`
 	BlockIndex    *uint32 `json:"blockindex,omitempty"`
 	CoinType      uint8   `json:"coin_type,omitempty"`
+	SKAAmountIn   string  `json:"ska_amountin,omitempty"`
 	SKAValue      string  `json:"ska_value,omitempty"`
 }
 
--- a/api/types/apitypes.go
+++ b/api/types/apitypes.go
@@ -88,20 +88,45 @@ type Tx struct {
 	Block         *BlockID `json:"block,omitempty"`
 }
 
-// Vin is an alias for dcrd's rpc/jsonrpc/types/v4.Vin type.
-type Vin = chainjson.Vin
+// ScriptSig models a transaction input's signature script.
+type ScriptSig struct {
+	Asm string `json:"asm"`
+	Hex string `json:"hex"`
+}
+
+// Vin models basic data about a transaction input.
+type Vin struct {
+	Coinbase     string     `json:"coinbase,omitempty"`
+	Stakebase    string     `json:"stakebase,omitempty"`
+	Treasurybase bool       `json:"treasurybase,omitempty"`
+	Txid         string     `json:"txid,omitempty"`
+	Vout         uint32     `json:"vout,omitempty"`
+	Tree         int8       `json:"tree,omitempty"`
+	ScriptSig    *ScriptSig `json:"scriptSig,omitempty"`
+	Sequence     uint32     `json:"sequence"`
+	AmountIn     float64    `json:"amountin,omitempty"`
+	BlockHeight  uint32     `json:"blockheight"`
+	BlockIndex   uint32     `json:"blockindex"`
+	CoinType     uint8      `json:"coin_type,omitempty"`
+	ValueInRaw   string     `json:"value_in_raw,omitempty"`
+}
 
 // TxShort models info about transaction TxID
 type TxShort struct {
-	TxID     string `json:"txid"`
-	Size     int32  `json:"size"`
-	Version  int32  `json:"version"`
-	Locktime uint32 `json:"locktime"`
-	Expiry   uint32 `json:"expiry"`
-	Vin      []Vin  `json:"vin"`
-	Vout     []Vout `json:"vout"`
-	Tree     int8   `json:"tree"`
-	Type     string `json:"type"`
+	TxID       string `json:"txid"`
+	Size       int32  `json:"size"`
+	Version    int32  `json:"version"`
+	Locktime   uint32 `json:"locktime"`
+	Expiry     uint32 `json:"expiry"`
+	Vin        []Vin  `json:"vin"`
+	Vout       []Vout `json:"vout"`
+	Tree       int8   `json:"tree"`
+	Type       string `json:"type"`
+	CoinType   uint8  `json:"coin_type"`
+	Total      string `json:"total"` // Atom string
+	Fee        string `json:"fee"`   // Atom string
+	FeeRate    string `json:"fee_rate,omitempty"`
+	FeeRateRaw string `json:"fee_rate_raw,omitempty"`
 }
 
 // AgendasInfo holds the high level details about an agenda.
@@ -121,12 +146,17 @@ type AgendaAPIResponse struct {
 
 // TrimmedTx models data to resemble to result of the decoderawtransaction RPC.
 type TrimmedTx struct {
-	TxID     string `json:"txid"`
-	Version  int32  `json:"version"`
-	Locktime uint32 `json:"locktime"`
-	Expiry   uint32 `json:"expiry"`
-	Vin      []Vin  `json:"vin"`
-	Vout     []Vout `json:"vout"`
+	TxID       string `json:"txid"`
+	Version    int32  `json:"version"`
+	Locktime   uint32 `json:"locktime"`
+	Expiry     uint32 `json:"expiry"`
+	Vin        []Vin  `json:"vin"`
+	Vout       []Vout `json:"vout"`
+	CoinType   uint8  `json:"coin_type"`
+	Total      string `json:"total"` // Atom string
+	Fee        string `json:"fee"`   // Atom string
+	FeeRate    string `json:"fee_rate,omitempty"`
+	FeeRateRaw string `json:"fee_rate_raw,omitempty"`
 }
 
 // Txns models the multi transaction post data structure
@@ -195,7 +225,8 @@ type VoutMined struct {
 
 // Vout defines a transaction output
 type Vout struct {
-	Value               float64      `json:"value"`
+	Value               float64      `json:"value"` // Deprecated: use ValueRaw
+	ValueRaw            string       `json:"value_raw"`
 	N                   uint32       `json:"n"`
 	Version             uint16       `json:"version"`
 	ScriptPubKeyDecoded ScriptPubKey `json:"scriptPubKey"`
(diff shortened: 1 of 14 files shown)


yanchenko-igor (approved):
Performance fix (de33496) — Eliminating N RPC calls for data the node already provides. This was the main finding during review.

edshav:
### Support for Unconfirmed Spenders and SKA Ingestion Path Fixes

This commit addresses three critical issues identified during live testing of the SKA transaction detail page:

#### 1. Mempool Spender Visibility (The "false" link bug)
**Problem**: When a confirmed transaction's output is spent by an unconfirmed transaction in the mempool, the explorer was failing to display the spending link, rendering a text `false` instead of a clickable hash. This occurred because the lookup logic only queried the PostgreSQL database, which does not contain mempool data.
**Solution**: 
- Added a `Spenders` lookup method to the mempool `DataCache`.
- Centralized spending detection in `ChainDB.SpendingTransactions`, which now performs a unified lookup across both the DB (confirmed) and mempool (unconfirmed).
- Updated `GetExplorerTx` to populate `SpendingTxns` and set the `Spent` flag by consulting this unified source, restoring unconfirmed spend links in the UI.

#### 2. SKA UTXO Ingestion Fix
**Problem**: The `updateUtxoCache` logic was discarding outputs with 0 value. While correct for VAR metadata, this caused the explorer to "forget" SKA outputs (which have 0 VAR value). Consequently, subsequent transactions spending those SKA outputs would fail to link back to their funding source during the initial sync.
**Solution**: Updated the caching logic to include outputs where `coin_type > 0`, ensuring all SKA UTXOs are indexed regardless of their VAR value.

#### 3. Address Analytics and Metrics
**Problem**: Several SQL queries in `addrstmts.go` were filtering by `value > 0`. Since SKA transactions carry 0 VAR value, they were being excluded from address history and balance calculations.
**Solution**: Updated filters to `(addresses.value > 0 OR addresses.coin_type > 0)`, ensuring SKA activity is correctly reflected in address analytics.

#### 4. Integrated Backend Performance Fix
These changes have been rebased and verified against @yanchenko-igor's performance improvements (`SKAAmountIn` integration), maintaining (1)$ RPC efficiency while adding full mempool awareness.

edshav:
### Fixed SKA Spending Status & UTXO Cache Discrepancy

This update resolves an issue where SKA transaction outputs were incorrectly displayed as **Spent (unlinked 'true')** even when they were unspent in the blockchain.

#### Root Cause
The explorer's in-memory **UTXO cache** was filtering out all transaction outputs with a VAR value of `0`. Because SKA transactions use `0` for VAR and store their balance in `ska_value`, they were being excluded from the cache during startup and sync. 
* This caused `GetExplorerTx` to report `Spent: true` (because the output was missing from the cache).
* Since the transaction was actually unspent, the database had no spender hash, resulting in the 'unlinked true' UI state.

#### Changes
1.  **Fixed Cache Warming Query**: Updated `SelectUTXOs` in `vinoutstmts.go` to include all outputs where `coin_type > 0` (SKA), even if the VAR value is `0`.
2.  **Tracking CoinType in Cache**: Updated the internal `utxoStore` and `UTXOData` structures to preserve `CoinType` in the in-memory map.
3.  **Corrected Live Ingestion**: Updated `updateUtxoCache` in `pgblockchain.go` to correctly add SKA outputs to the cache as they are mined.
4.  **Fixed Address History Flattening**: Patched `flattenAddressRows` to stop discarding SKA rows with `0` VAR value, ensuring correct address history and balance tracking.

These fixes ensure that SKA transactions correctly display as **Unspent** (or link to their actual spenders) after a restart or during live synchronization.

yanchenko-igor (approved):

PR #125, opened 2026-04-22
fix: display VAR supply with full precision and trim trailing zeros

### LINKED ISSUE #136: Implement high-precision supply logic
## Tasks
- [ ] Return full precision values
- [ ] Trim trailing zeros
- [ ] Ensure no rounding
- [ ] Provide Issued and Burned values

## Validation
- [ ] Matches spec examples

## Expected Result
Accurate financial-grade supply data.

Part of #129

yanchenko-igor:
VAR circulating supply was displaying rounded to whole numbers (0 decimals). Changed to show full 8-decimal precision matching SKA display pattern.

Changes:
- home_supply.tmpl: numPlaces 0 → 8 for VAR supply
- supply_controller.js: decimalPlaces 0 → 8 for JS live updates
- float64Formatting: trim trailing zeros like skaDecimalParts

Before: 1,624,749
After: 1,624,749.8 (without trailing zeros)

### DIFF
--- a/cmd/dcrdata/internal/explorer/templates.go
+++ b/cmd/dcrdata/internal/explorer/templates.go
@@ -142,9 +142,6 @@ func float64Formatting(v float64, numPlaces int, useCommas bool, boldNumPlaces .
 	pow := math.Pow(10, float64(numPlaces))
 	formattedVal := math.Round(v*pow) / pow
 	clipped := fmt.Sprintf("%."+strconv.Itoa(numPlaces)+"f", formattedVal)
-	oldLength := len(clipped)
-	clipped = strings.TrimRight(clipped, "0")
-	trailingZeros := strings.Repeat("0", oldLength-len(clipped))
 	valueChunks := strings.Split(clipped, ".")
 	integer := valueChunks[0]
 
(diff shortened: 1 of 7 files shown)


edshav (commented):
## Review

The three commits are well-structured: core fix first, then JS helper cleanup, then removing the now-unused variable. Clean progression.

### Issues

**`supply_controller.js` — Float precision risk**

```js
parseInt(ex.var_coin_supply.circulating) / 1e8
```
`parseInt` + `/1e8` can silently lose precision for atom values above `Number.MAX_SAFE_INTEGER` (~9 quadrillion atoms = ~90M VAR). The codebase already has `humanize.formatCoinAtomsFull(atomStr, 0)` which handles this correctly via BigInt. Worth switching to that.

**`templates.go` — Scope of change is broader than described**

`float64Formatting` now always trims trailing zeros and always returns `""` in the trailing-zeros slot. This affects every caller, not just VAR supply. The PR description only mentions VAR. Worth noting in the description whether this is intentional for all callers, or if it should be scoped.

### Looks Good

- `home_supply.tmpl` and `supply_controller.js` are consistent (`0 → 8` in both).
- `humanize_helper.js` is clean — `trailingZeros` fully removed across all three commits, no dangling references.
- The trailing-zero trimming in `float64Formatting` now matches `skaDecimalParts` behavior, which is a reasonable consistency win.

edshav:
Why did you decide to remove the trailing zeros?

edshav:
> Why did you decide to remove the trailing zeros?

Ah, its in the issue 👍

edshav (approved):
All feedback addressed. The two new commits correctly fix the BigInt precision issue and the comma-stripping edge case.

I've pushed an additional commit on this branch: coerces all `decimalParts` call sites in `mining_controller`, `mempool_controller`, and `voting_controller` to pass `String()` explicitly, and added a `String(v)` guard inside `decimalParts` itself as a safety net. All pre-commit checks (prettier, eslint, vitest) pass.

LGTM.

PR #142, opened 2026-04-22
Fix/pow ska rewards extraction

yanchenko-igor:
## Summary
- **Root cause**: SKA PoW rewards were being extracted from coinbase, but coinbase is always empty for SKA. The actual SKA fees come from regular transaction inputs minus outputs.
- **Fix**: Extract SKA fees from regular transactions in `BlockSKAFees()` (input SKA - output SKA = miner fee), and use RPC in fallback search to compute fees when cached data is unavailable.
## Changes
- Add `BlockSKAFees()` function that calculates transaction fees from regular (non-stake) transactions
- Add `GetBlockSKAFees()` method on ChainDB that uses RPC to fetch raw block and compute SKA fees
- Add fallback logic in explorer UI that searches backwards (up to 4320 blocks) using RPC to find last block with SKA transactions
## Testing
- Unit tests added for `BlockSKAFees()` covering both SKA transactions and empty blocks
- Live tested: fallback finds block 30067 with ~2 SKA fee
## Files Changed
- `blockdata/blockdata.go`: Add `BlockSKAFees()` function
- `blockdata/blockdata_test.go`: Add tests
- `cmd/dcrdata/internal/explorer/explorer.go`: Fallback logic with RPC
- `db/dcrpg/pgblockchain.go`: Add `GetBlockSKAFees()` method

### DIFF
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -259,9 +259,9 @@ func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataB
 		extrainfo.SSFeeTotalsByCoin = ssfee
 	}
 
-	// Extract per-SKA-type PoW mining reward amounts from the coinbase.
-	if skaRewards := BlockSKAPoWRewards(msgBlock); len(skaRewards) > 0 {
-		extrainfo.SKAPoWRewards = skaRewards
+	// Extract SKA fees from regular transactions (goes to miner as PoW reward).
+	if skaFees := BlockSKAFees(msgBlock); len(skaFees) > 0 {
+		extrainfo.SKAPoWRewards = skaFees
 	}
 
 	return blockdata, feeInfoBlock, blockHeaderResults, extrainfo, msgBlock, err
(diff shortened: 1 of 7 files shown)


edshav (commented):
## Summary

This PR correctly identifies and fixes a real bug: SKA PoW rewards were being read from the coinbase transaction, but SKA fees actually flow through regular transaction inputs/outputs. The fix is architecturally sound and the approach (input SKA − output SKA = miner fee) is the right model. The fallback path in the explorer is also meaningfully simplified. Overall this is a good change, but there are a few issues worth addressing.

---

## Issues

**1. `BlockSKAFees` hardcodes `ct := uint8(1)` — this is the most significant bug in the PR.**

```go
ct := uint8(1)
if skaFees[ct] == nil {
    skaFees[ct] = fee
} else {
    skaFees[ct].Add(skaFees[ct], fee)
}
```

The function is supposed to handle multiple SKA coin types (1–255), but the fee is always bucketed under type `1`. Compare with `ExtractSKARewardsFromCoinbase`, which correctly derives `ct` from `txout.CoinType`. The fix should derive the coin type from the input's `SKAValueIn` context — or, since inputs don't carry a coin type directly, from the corresponding outputs. This needs a rethink: you'd need to match inputs to their coin type via the outputs of the referenced transactions, or at minimum derive it from the outputs of the current tx that are SKA-typed.

**2. Test coverage misses the hardcoded `ct` bug.**

`TestBlockSKAFees_WithSKATransaction` only exercises coin type 1, so it doesn't catch the hardcoded `ct` issue. A test with multiple SKA coin types would have caught this.

**3. Dead code / deprecation plan missing.**

The `DEPRECATED` comment on `BlockSKAPoWRewards` is good, but the old function and `ExtractSKARewardsFromCoinbase` are still exported with no removal plan noted. If they're truly dead code paths now, a follow-up issue or TODO would help avoid confusion.

**4. Minor: redundant zero-initialization in `BlockSKAFees`.**

`inputSKA.SetInt64(0)` and `outputSKA.SetInt64(0)` are redundant — `big.Int` zero-values are already zero. Not a bug, just noise.

**5. RPC-per-block in fallback loop — worth a comment.**

`GetBlockSKAFees` makes an RPC call per block in the fallback loop (up to 4320 calls). This was true of the old path too, but since the new path is more explicit about using RPC, a comment noting the performance implication would be helpful.

---

## What's good

- Root cause analysis is correct and well-explained in the PR description.
- `GetBlockSKAFees` cleanly delegates to `BlockSKAFees`, keeping the DB layer thin.
- Removing the coinbase-address-matching aggregation from `GetExplorerBlock` is a meaningful simplification — that code was complex and fragile.
- The mock pattern in `explorer_test.go` using a package-level `mockGetBlockSKAFeesResult` with `defer` cleanup is safe enough for now.
- Unit tests for the happy path and no-SKA case are present.

---

The hardcoded `ct := uint8(1)` is a blocker — it silently drops fees for any SKA coin type other than 1. Everything else is minor.

edshav (approved):
All blockers and minors from the previous review are addressed:

- **Hardcoded `ct := uint8(1)` fixed** — coin type is now derived from the first SKA output per transaction. The single-coin assumption is documented in the comment.
- **Test coverage** — `TestBlockSKAFees_MultipleCoinTypes` covers types 1, 2, and 5. Solid.
- **Deprecation/TODO** — both `ExtractSKARewardsFromCoinbase` and `BlockSKAPoWRewards` have `DEPRECATED` + `TODO: Remove in next minor version` comments.
- **Redundant zero-init** — `SetInt64(0)` calls removed.
- **RPC performance note** — the 4320-call warning comment is in place.
- **SKA naming** — `SKA-` → `SKA` in templates, consistent.

One minor nit: test style is slightly inconsistent between `TestBlockSKAFees_WithSKATransaction` (appends extra inputs) and `TestBlockSKAFees_MultipleCoinTypes` (sets `TxIn[0].SKAValueIn` directly). Not a blocker — the new style is actually cleaner.

Good to merge.

Issue #185, opened 2026-04-28
[UI] Update 'Current Mempool' card to show per-coin totals


edshav:
The 'Total Sent' field must display a breakdown by coin: VAR first, then SKA types in ascending order. Only coins present in the current mempool are shown. If a coin has no transactions, it is not displayed.

VAR amounts: 8 decimal places. SKA amounts: 18 decimal places.

The card must reflect live mempool state without page reload.

Part of #183

PR #192, opened 2026-04-28
feat: mempool per-coin SKA data for summary cards, transactions table, and WebSocket

### LINKED ISSUE #184: [DATA] Expose per-coin mempool aggregates and per-transaction coin type
The frontend requires the following data to render the updated mempool page:

**Summary cards:**
- Total sent amount per coin (VAR + each SKA present in mempool), for the 'Current Mempool' card
- Per-coin transaction counts and totals broken down by type (Regular, Ticket, Vote, Revocation) for the 'Transactions' card. SKA coins expose Regular only.

**Regular transactions table:**
- Each transaction entry must include its coin type (VAR or SKA-n) and amount
- VAR amounts: up to 8 decimal places. SKA amounts: up to 18 decimal places, string-based, no float64 coercion

**WebSocket:**
- All of the above must be available in real-time update payloads, not only at page load

Reference: `wiki/code-analysis/mempool/flow.full.md` §3, §5, §8

Part of #183

yanchenko-igor:
Summary
This PR extends the mempool backend and frontend to expose per-coin transaction data for VAR and each active SKA coin type. It adds per-type tx counts to CoinStats, propagates SKA coin identity through TrimmedTxInfo, and updates the mempool page template and JS controller to render coin-aware amounts at full precision (18dp for SKA, 8dp for VAR, no float64 coercion).
Changes
Backend
mempool/monitor.go — Per-type counter increments in TxHandler
When a new transaction enters the mempool, the per-coin CoinStats now tracks how many Regular, Ticket, Vote, and Revocation transactions exist for each coin type (VAR key=0, SKA-n key=n). Increments happen alongside the existing amount aggregation in both the VAR branch (empty SKATotals) and the SKA loop.
mempool/collector.go — Per-type counters in ParseTxns (full inventory rebuild)
The same per-type counting is applied when the full mempool inventory is rebuilt (e.g., after a new block). An internal coinAccum struct accumulates counts per coin type; the final map is converted to MempoolCoinStats with all four type counters populated.
explorer/types/explorertypes.go — Type extensions and TrimMempoolTx fix
- MempoolCoinStats gains RegularCount, TicketCount, VoteCount, RevokeCount (all int, all zero for SKA).
- TrimmedMempoolInfo gains CoinStats field, copied in Trim().
- TrimMempoolTx now sets TxBasic.CoinType and TxBasic.SKASent from the SKATotals map so TrimmedTxInfo carries the coin identity. The duplicate code block from the original implementation is removed.
cmd/dcrdata/internal/explorer/templates.go — New helpers
- coinSymbol(coinType uint8) — existing function, registered in makeTemplateFuncMap.
- primaryCoinType(skaTotals map[uint8]string) uint8 — extracts coin type from a SKATotals map (for use with raw MempoolTx which lacks a CoinType field). Registered as primaryCoinType.
Frontend (minimal demo)
mempool.tmpl — "Current Mempool" card and Regular transactions table
- "Total Sent" iterates .CoinStats via range $ct, $stats, rendering each coin's label via coinSymbol and total via formatCoinAtoms. The layout is simplified to a clean stacked display.
- Regular transactions table adds Coin and Amount columns. Uses primaryCoinType to derive coin type from SKATotals, then formatCoinAtomsFull for SKA amounts. Falls back to existing float64AsDecimalParts for VAR.
js/helpers/humanize_helper.js — coinSymbol for JS
Added coinSymbol function and exported it on the humanize object for use by the controller.
js/controllers/mempool_controller.js — Real-time tx row template
Updated txTableRow to check tx.ska_totals, derive coin type, and use humanize.formatCoinAtomsFull (JS BigInt → exact 18dp for SKA) or decimalParts (8dp for VAR). This fixes the first row of the table, which is prepended by the JS controller on new transaction notifications — not by the Go template.
WebSocket payload
No structural changes. The existing sigMempoolUpdate path carries the full MempoolInfo which now includes CoinStats (with per-type breakdown) and each MempoolTx already carries SKATotals (mapuint8string). The frontend's JS already subscribes to newtxs and mempool events.
Precision guarantees
- All SKA amounts are stored and transmitted as decimal atom strings (big.Int.String()), never as float64.
- addAtomStrings(..., isBig=true) uses big.Int for SKA arithmetic.
- JS uses BigInt and explicit decimal division without Math.pow coercion.
- VAR uses int64 atom strings, divided by 1e8 for display.
Testing
Builds clean with go build ./.... No new test files introduced.
Checklist
- [x] SKA amounts use big.Int / BigInt, not float64
- [x] VAR amounts use int64 atoms, not float64
- [x] Per-type counts track Regular/Ticket/Vote/Revocation per coin
- [x] TrimmedTxInfo preserves CoinType and SKASent
- [x] WebSocket delivers CoinStats and SKATotals via existing sigMempoolUpdate path
- [x] Real-time JS txTableRow handles SKA amounts

### DIFF
--- a/explorer/types/explorertypes.go
+++ b/explorer/types/explorertypes.go
@@ -665,9 +665,9 @@ type TrimmedMempoolInfo struct {
 	Total          float64
 	Time           int64
 	Fees           float64
-	CoinFills      []CoinFillData `json:"coin_fills,omitempty"`
-	TotalFillRatio float64        `json:"total_fill_ratio"`
-	ActiveSKACount int            `json:"active_ska_count"`
+	CoinFills      []CoinFillData             `json:"coin_fills,omitempty"`
+	TotalFillRatio float64                    `json:"total_fill_ratio"`
+	ActiveSKACount int                        `json:"active_ska_count"`
 	CoinStats      map[uint8]MempoolCoinStats `json:"coin_stats,omitempty"`
 }
 
@@ -862,13 +862,25 @@ func TrimMempoolTxs(txs []MempoolTx) []*TrimmedTxInfo {
 	return trimmedTxs
 }
 
-// TrimMempoolTx converts the input []MempoolTx to a []*TrimmedTxInfo.
+// primaryCoinType extracts the coin type from a mempool transaction's SKATotals map.
+// INVARIANT: A mempool transaction contains outputs of exactly one coin type
+// (either VAR with no SKATotals, or exactly one SKA coin type). This function
+// returns the coin type found, or 0 for VAR (empty SKATotals map).
+func primaryCoinType(tx *MempoolTx) uint8 {
+	for ct := range tx.SKATotals {
+		return ct
+	}
+	return 0
+}
+
+// TrimMempoolTx converts the input MempoolTx to a TrimmedTxInfo for display.
 func TrimMempoolTx(tx *MempoolTx) (trimmedTx *TrimmedTxInfo) {
 	fee, _ := dcrutil.NewAmount(tx.Fees) // non-nil error returns 0 fee
 	var feeRate dcrutil.Amount
 	if tx.Size > 0 {
 		feeRate = fee / dcrutil.Amount(int64(tx.Size))
 	}
+	coinType := primaryCoinType(tx)
 	txBasic := &TxBasic{
 		TxID:          tx.TxID,
 		Type:          tx.Type,
@@ -878,13 +890,8 @@ func TrimMempoolTx(tx *MempoolTx) (trimmedTx *TrimmedTxInfo) {
 		Fee:           fee,
 		FeeRate:       feeRate,
 		VoteInfo:      tx.VoteInfo,
-		CoinType: func() uint8 {
-			for ct := range tx.SKATotals {
-				return ct
-			}
-			return 0
-		}(),
-		SKASent: tx.SKATotals,
+		CoinType:      coinType,
+		SKASent:       tx.SKATotals,
 		// TreasuryBase and Coinbase are not in mempool
 	}
 	var voteValid bool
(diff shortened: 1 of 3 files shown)


edshav (changes requested):
**Overall:** The backend data changes are solid — per-type counts are correctly accumulated in both the full rebuild (`collector.go`) and the incremental path (`monitor.go`), and the precision story for SKA amounts is well-handled throughout.

However, this PR mixes data and UI concerns. Issue #184 is scoped to data only; the UI work belongs to issues #185–189. The following files should be reverted:

- `cmd/dcrdata/views/mempool.tmpl`
- `cmd/dcrdata/public/js/controllers/mempool_controller.js`
- `cmd/dcrdata/public/js/helpers/humanize_helper.js`
- `cmd/dcrdata/internal/explorer/templates.go` — `primaryCoinType` and the `coinSymbol` registration are only needed to support the template changes above, so they should be deferred along with the UI files.

---

**Issue — misleading naming in `TrimMempoolTx` (explorertypes.go)**

The inline function used to extract coin type from `SKATotals`:

```go
CoinType: func() uint8 {
    for ct := range tx.SKATotals {
        return ct
    }
    return 0
}(),
```

The pattern of iterating a map to get "the" entry implies there could be multiple. Since a transaction is guaranteed to have exactly one coin, this should be a named helper with a comment documenting that invariant, rather than an anonymous loop that looks like it's picking the first of many.

---

**Minor notes**

- `CoinStats` propagation through `Trim()` is clean and correct.
- The `TrimmedTxInfo` now correctly carries `CoinType` and `SKASent` — good fix.
- Removing the duplicate code block from the original `TrimMempoolTx` is a welcome cleanup.

edshav (approved):

PR #199, opened 2026-04-28
Feature/ska reward block link

### LINKED ISSUE #197: [DATA] Add a block height to a template context
We need to make a SKA{n}/Vote value as a link to the corresponding block e.g. `/block/34014`.

Add a block height for the SKA{n}/Vote value to the `cmd/dcrdata/views/home_voting.tmpl` template context. 

Image

### LINKED ISSUE #198: [DATA] Add a block height to a template context
We need to make a `PoW SKA Reward` value as a link to the corresponding block e.g. `/block/34014`

Add the corresponding block height for the `PoW SKA Reward` value to the `cmd/dcrdata/views/home_mining.tmpl` template context.

Image

yanchenko-igor:
Makes the PoW SKA Reward and Vote SKA Reward sections on the home page 
clickable, linking to the block that produced each reward value.

Closes #197 #198

### DIFF
--- a/cmd/dcrdata/internal/explorer/explorer.go
+++ b/cmd/dcrdata/internal/explorer/explorer.go
@@ -666,9 +666,11 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
 	// Find the latest block with SKA fee data for PerBlock calculation.
 	var latestSkaBlock *apitypes.BlockDataBasic
 	var latestSkaVoters int64
+	var voteRewardsBlockHeight int64
 
 	if len(blockData.ExtraInfo.SSFeeTotalsByCoin) > 0 {
 		// Use current block data directly to avoid redundant DB query
+		voteRewardsBlockHeight = int64(blockData.Header.Height)
 		latestSkaVoters = int64(blockData.Header.Voters)
 		latestSkaBlock = &apitypes.BlockDataBasic{
 			SSFeeTotalsByCoin: blockData.ExtraInfo.SSFeeTotalsByCoin,
@@ -678,6 +680,7 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
 		for i := len(sum30) - 1; i >= 0; i-- {
 			if len(sum30[i].SSFeeTotalsByCoin) > 0 {
 				latestSkaBlock = sum30[i]
+				voteRewardsBlockHeight = int64(sum30[i].Height)
 				// Get full block info to retrieve the voters count for this specific block
 				if bInfo := exp.dataSource.GetExplorerBlock(ctx, latestSkaBlock.Hash); bInfo != nil {
 					latestSkaVoters = int64(bInfo.BlockBasic.Voters)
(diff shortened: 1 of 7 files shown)


edshav:
Fixed the UI issues in commit 7eec9052:

- **`home_voting.tmpl`**: replaced the split `{{if .BlockHeight}}<a>` / `{{if .BlockHeight}}</a>` pattern with a proper self-contained conditional block (`{{if .BlockHeight}}<a>...</a>{{else}}<div>...</div>{{end}}`). The old pattern was fragile — mismatched tags across separate `{{if}}` blocks are easy to break on future template edits.
- **`home_mining.tmpl`**: guarded the initial server-side render with `{{if .BlockHeight}}` so a zero height never produces a `/block/0` link. Falls back to a plain `<div>` with "per last block" text, consistent with the JS live-update path.
- **`mining_controller.js`**: removed the unused `'blockHeight'` entry from the Stimulus `targets` array. The controller already queries `[data-block-height]` directly on cloned template nodes — no target descriptor needed.

edshav:
Two backend issues worth addressing before merge:

**1. `pubsubhub.go` — vote reward block height always uses current block**

```go
voteRewardsBlockHeight := int64(blockData.Header.Height)
```

This is set unconditionally to the current block height, then stamped on every `SKAVoteReward` entry regardless of whether the current block actually contains SKA fee data. The explorer's `Store` path in `explorer.go` does the right thing — it walks `sum30` backwards to find the most recent block that has SKA fee data and uses *that* block's height. `pubsubhub.go` skips that walk, so the vote reward link delivered via WebSocket can point to a block with no SKA fees.

Fix: mirror the `explorer.go` logic — only set `voteRewardsBlockHeight` when `blockData.ExtraInfo.SSFeeTotalsByCoin` is non-empty for the current block, and otherwise find the latest block in `sum30` that has fee data (same loop that already runs to compute `perBlock`).

**2. `home_voting.tmpl` / `voting_controller.js` — inconsistent no-height rendering between initial load and live update**

On initial server-side render, when `BlockHeight == 0` the reward is wrapped in a plain `<div>` (no link). On a live WebSocket update with no height, `voting_controller.js` still renders the `<a data-block-height>` element from the template but leaves `href` unset — so the DOM contains an `<a>` with no `href`, which is technically non-interactive but structurally different from the initial render. Same applies to the mining template/controller pair.

This is low severity (browsers treat `<a>` without `href` as non-interactive), but if it matters for consistency the JS templates could conditionally swap the wrapper element, or the `<a>` could be replaced with a `<span>` when height is absent.

edshav:
Issue 2 (`<a>` vs `<div>` inconsistency) has been fixed in commit 3ae6a532.

Both `home_voting.tmpl` and `home_mining.tmpl` now always render `<a>` for each reward entry — `href` is conditionally set only when `BlockHeight > 0`, matching exactly what the JS controllers do on live WS updates. The `<a>`/`<div>` branching is gone.

Issue 1 (`pubsubhub.go` vote reward block height) is still open for the backend.

edshav:
Reviewed the two new commits.

**`365d624e` — fix(pubsubhub): only set vote reward block height when current block has SKA fees**

Correct fix for issue 1. One nuance worth knowing: the comment says this matches `explorer.go` behavior, but `explorer.go` actually goes a step further — it walks `sum30` backwards to find the most recent block that has SKA fees and stamps *that* height, so the link is present even when the current block has no fees. `pubsubhub.go` leaves height as `0` in that case, so the vote reward link will disappear on a WS update for a fee-free block and reappear on the next block with fees. Not a correctness bug, but a minor UX difference between initial load and live updates.

**`982cdb6b` — fix: remove block height from SKA reward display text**

This one is a step back. The label now always reads "per last block" even when the `<a>` has a valid `href` pointing to a specific block — so the element is a clickable link but the text doesn't tell the user which block it links to. The original conditional label (`per block {height}` when height is known, `per last block` otherwise) was the right approach. Suggest reverting this commit or restoring the conditional text in both the template and the JS controller.

edshav (commented):
Comment posted above

edshav (approved):

PR #202, opened 2026-04-29
Feature/add last block reward data

### LINKED ISSUE #203: PoW VAR Reward add details
PoW VAR Reward

Добавить расшифровку под значением:

строка субсидии
строка fee
Subsidy: 3.20000000
Fee: 0.00364000

yanchenko-igor:
Summary
Display total PoW VAR reward (subsidy + fees) for the latest block on the home page, with a breakdown showing subsidy and fees separately.
Changes
Backend (explorer/types, explorer.go, pubsubhub.go, explorerroutes.go):
- Added LBlockSubsidy, MiningFee, LBlockTotal, LBlockTotalAtoms fields to HomeInfo to store total reward and fee data
- Compute total reward = subsidy (~16 VAR) + mining fees from latest block
- Both web UI and websocket paths populate these fields
Frontend (home_mining.tmpl, mining_controller.js):
- Main display now shows total reward (LBlockTotalAtoms) instead of predicted next block subsidy (NBlockSubsidy.PoW)
- Added breakdown lines showing:
  - Subsidy: 16.00000000 - the block subsidy  
  - Fee: 0.00767328 - mining fees from transactions
- Fixed JS operator precedence bug in fallback logic
- Uses toFixed(8) to preserve trailing zeros on websocket updates
Before/After
- Before: Showed predicted next block subsidy (~16 VAR) without fees
- After: Shows total reward (subsidy + fees, ~16.01 VAR) with breakdown
Testing
- Verified on initial page load and after websocket updates
- Values consistent between server-side render and client-side updates

### DIFF
--- a/api/types/apitypes.go
+++ b/api/types/apitypes.go
@@ -796,6 +796,7 @@ type BlockExplorerExtraInfo struct {
 	TxLen            int                              `json:"tx"`
 	CoinSupply       int64                            `json:"coin_supply"`
 	NextBlockSubsidy *chainjson.GetBlockSubsidyResult `json:"next_block_subsidy"`
+	MiningFeeAtoms   int64                            `json:"mining_fee_atoms"`
 	// CoinAmounts holds per-coin totals (VAR key=0, SKAn key=n) as decimal atom strings.
 	CoinAmounts map[uint8]string `json:"coin_amounts,omitempty"`
 	// CoinTxStats holds per-coin tx count and size (key 0=VAR, 1-255=SKAn).
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -239,6 +239,8 @@ func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataB
 	if feeInfoBlock == nil {
 		log.Error("FeeInfoBlock failed")
 	}
+	// Compute mining fee (total fees from all transactions in block)
+	miningFee := computeMiningFee(msgBlock)
 
 	// Work/Stake difficulty
 	diff := txhelpers.GetDifficultyRatio(header.Bits, t.netParams)
@@ -258,6 +260,7 @@ func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataB
 		TxLen:            txLen,
 		CoinSupply:       int64(coinSupply),
 		NextBlockSubsidy: nbSubsidy,
+		MiningFeeAtoms:   miningFee,
 	}
 
 	// Accumulate per-coin output totals from all transactions in the block.
@@ -584,4 +587,23 @@ func BlockSKAFees(msgBlock *wire.MsgBlock) map[uint8]string {
 		out[k] = v.String()
 	}
 	return out
+}
+
+// computeMiningFee calculates total mining fees from a block (sum of all tx fees).
+func computeMiningFee(msgBlock *wire.MsgBlock) int64 {
+	var totalFee int64
+	for i, tx := range msgBlock.Transactions {
+		if i == 0 {
+			continue // skip coinbase
+		}
+		var in, out int64
+		for _, vin := range tx.TxIn {
+			in += vin.ValueIn
+		}
+		for _, vout := range tx.TxOut {
+			out += vout.Value
+		}
+		totalFee += in - out
+	}
+	return totalFee
 }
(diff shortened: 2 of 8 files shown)


edshav (commented):
**Summary**

This PR adds total PoW reward display (subsidy + fees) to the home page, replacing the predicted next-block subsidy with the actual last-block total. The intent is solid and the implementation mostly works, but there are a few issues worth addressing.

---

**Issues**

**1. Hardcoded fee estimate in `explorerroutes.go` (medium)**

```go
homeInfo.LBlockTotalAtoms = homeInfo.NBlockSubsidy.PoW + 10000000 // ~0.1 VAR estimate
homeInfo.LBlockTotal = dcrutil.Amount(homeInfo.NBlockSubsidy.PoW).ToCoin() + 0.1
homeInfo.MiningFee = 0.1 // estimate
```

The fallback path in `Home()` uses a hardcoded `0.1 VAR` fee estimate when `LBlockTotalAtoms == 0`. This will show incorrect data on first load if `GetExplorerBlock` hasn't populated the value yet. It would be better to either leave the fee as 0 (showing only subsidy) or skip the fallback entirely and let the template handle a zero value gracefully.

**2. Duplicate `GetExplorerBlock` call (performance)**

Both `explorer.go` (`Store`) and `pubsubhub.go` (`Store`) independently call `GetExplorerBlock` to fetch mining fee data. This means every new block triggers two separate lookups for the same block. Consider computing the fee once and passing it through `BlockData` or a shared struct.

**3. Float precision risk in `LBlockTotalAtoms` calculation**

```go
p.HomeInfo.LBlockTotalAtoms = p.HomeInfo.NBlockSubsidy.PoW + int64(blockInfo.MiningFee*1e8)
```

`MiningFee` is a `float64`. Multiplying by `1e8` and casting to `int64` can lose precision for certain fee values. The safer approach is to store `MiningFee` in atoms (int64) from the source and avoid the float conversion entirely.

**4. Removed fiat conversion display in template**

The fiat conversion line was removed from the template without explanation. The JS controller still updates `powConverted`, so the data is computed but never shown. Was this intentional? If so, it should be noted in the PR description.

**5. `LBlockSubsidy` field added but never populated**

`HomeInfo` gains a `LBlockSubsidy BlockSubsidy` field in `explorertypes.go`, but nothing in the diff writes to it. Either populate it or remove it to avoid dead fields in the struct.

---

**Minor**

- The PR title uses a branch name format — a more descriptive title like "Show total PoW reward (subsidy + fees) on home page" would be clearer.
- Debug `log.Debugf` calls are fine for now but worth noting they'll appear in production logs at debug level.

---

**Overall**

The feature works for the happy path, but the hardcoded fee fallback and float-to-int64 conversion are the two things I'd want fixed before merging. The duplicate `GetExplorerBlock` call is a nice-to-have cleanup.

edshav (approved):
Good updates. Here's my draft review — not posted yet.

---

**PR #202 — Follow-up review (updated commits)**

The author addressed all five issues from the previous review. Here's how each landed:

**✅ Issue 1 — Hardcoded fee estimate: Fixed**
The `0.1 VAR` hardcoded estimate is gone. The fallback in `explorerroutes.go` now correctly uses `MiningFeeAtoms = 0` and sets `LBlockTotalAtoms` to just the subsidy when fees aren't available yet. Clean.

**✅ Issue 2 — Duplicate `GetExplorerBlock` call: Fixed**
Both `explorer.go` and `pubsubhub.go` now read `MiningFeeAtoms` directly from `blockData.ExtraInfo` instead of making a second `GetExplorerBlock` lookup. The fee is computed once in the collector (`blockdata.go`) via the new `computeMiningFee` function.

**✅ Issue 3 — Float precision risk: Fixed**
`MiningFee float64` is replaced with `MiningFeeAtoms int64` throughout. All arithmetic is now integer-based, eliminating the float-to-int64 precision issue.

**✅ Issue 4 — Removed fiat conversion: Acknowledged**
The template still omits the fiat conversion line, which appears intentional. The JS controller still computes and updates `powConverted` correctly for when the target element is present elsewhere.

**✅ Issue 5 — Unused `LBlockSubsidy` field: Fixed**
The dead `LBlockSubsidy BlockSubsidy` field has been removed from `HomeInfo`.

---

**New issue: binary committed to repo**

```
cmd/dcrdata/dcrdata  (new file, binary, 100755)
```

A compiled binary was accidentally committed. This should not be in version control — it belongs in `.gitignore`. Please remove it before merging.

---

**Minor**
- `computeMiningFee` in `blockdata.go` is clean and correct. The coinbase skip (`i == 0`) is the right approach.
- The `log.Debugf` calls remain in both `explorer.go` and `pubsubhub.go`. Fine for now, but worth a cleanup pass before a production release.

---

**Overall**

All previous feedback was addressed well. The only blocker is the committed binary at `cmd/dcrdata/dcrdata` — remove that and this is good to merge.

---

Want me to post this?

PR #212, opened 2026-04-30
fix: calculate correct SKA vote rewards by subtracting inputs

### LINKED ISSUE #206: [DATA] Vote SKA Reward
- Пересчитать Числовое значение SKA{n}/VOTE
- добавить доходность SKA{n}/VAR per year

yanchenko-igor:
Summary
This PR fixes the overestimation of SKA staking (voting) rewards and implements a more accurate, ticket-specific annual percentage yield (APY) calculation for the home page. It also cleans up the UI by removing the 30-day average for SKA rewards.
Changes
1. Corrected Reward Calculation
- txhelpers/ssfee.go: Updated BlockSSFeeTotals to calculate the actual reward as the difference between total SKA outputs and total SKA inputs (outputs - inputs). Previously, it only summed outputs, which incorrectly included consolidated funds (dust UTXOs) as rewards.
2. Accurate Per-Year APY Implementation
- cmd/dcrdata/internal/explorer/explorer.go: Replaced the simple block-average for PerYear with a precise calculation based on the latest reward block:
    - Formula: $\text{APY} = \frac{1}{N} \sum_{i=1}^{N} \left( \frac{365 \times (\frac{\text{Reward Per Ticket}}{\text{Ticket Price}})}{\text{Ticket Age in Blocks}} \right)$
    - The system now fetches the specific purchase price and purchase height for every ticket that voted in the last reward block.
    - Rewards are correctly converted from atoms to coins (divided by $10^{18}$) to ensure the ratio is expressed in SKA coins per VAR coin.
- db/dbtypes/types.go: Added VoteTicketData struct to transport purchase price and height.
- db/dcrpg/internal/stakestmts.go & db/dcrpg/queries.go: Added SelectVoteTicketDataByBlock to join votes and tickets tables for efficient data retrieval.
- db/dcrpg/pgblockchain.go: Implemented GetVoteTicketDataByBlock method.
3. UI & API Cleanup
- explorer/types/explorertypes.go: Removed Per30Days from SKAVoteReward as it is no longer used.
- cmd/dcrdata/views/home_voting.tmpl: Removed the "per 30 days" line from the SKA reward display.
- pubsub/pubsubhub.go: Removed obsolete 30-day average calculations for SKA rewards to reduce overhead.
Verification
- Correctness: Verified that the APY no longer shows astronomical values (fixed the $10^{18}$ scaling issue).
- Performance: The new lookup uses an indexed join on ticket_hash, ensuring fast retrieval even for blocks with many voters.
- Stability: All existing tests in home_voting_template_test.go and other packages pass.
- Linting: golangci-lint reports 0 issues.

### DIFF
--- a/cmd/dcrdata/internal/explorer/explorer.go
+++ b/cmd/dcrdata/internal/explorer/explorer.go
@@ -670,54 +670,48 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
 		coinTypes[ct] = struct{}{}
 	}
 
-	// Find the latest block with SKA fee data for PerBlock calculation.
-	var latestSkaBlock *apitypes.BlockDataBasic
-	var latestSkaVoters int64
-	var voteRewardsBlockHeight int64
-
-	if len(blockData.ExtraInfo.SSFeeTotalsByCoin) > 0 {
-		// Use current block data directly to avoid redundant DB query
-		voteRewardsBlockHeight = int64(blockData.Header.Height)
-		latestSkaVoters = int64(blockData.Header.Voters)
-		latestSkaBlock = &apitypes.BlockDataBasic{
-			SSFeeTotalsByCoin: blockData.ExtraInfo.SSFeeTotalsByCoin,
-		}
-	} else {
-		// Fallback: Search backwards through historical summaries
-		for i := len(sum30) - 1; i >= 0; i-- {
-			if len(sum30[i].SSFeeTotalsByCoin) > 0 {
-				latestSkaBlock = sum30[i]
-				voteRewardsBlockHeight = int64(sum30[i].Height)
-				// Get full block info to retrieve the voters count for this specific block
-				if bInfo := exp.dataSource.GetExplorerBlock(ctx, latestSkaBlock.Hash); bInfo != nil {
-					latestSkaVoters = int64(bInfo.BlockBasic.Voters)
-				}
-				break
-			}
-		}
-	}
-
 	if len(coinTypes) > 0 {
 		sum30S := toSummaries(sum30)
 		sumYearS := toSummaries(sumYear)
 		rewards := make([]types.SKAVoteReward, 0, len(coinTypes))
 		for ct := range coinTypes {
 			var perBlock string
-			if latestSkaBlock != nil {
-				if totalStr, ok := latestSkaBlock.SSFeeTotalsByCoin[ct]; ok {
-					if total, parsed := new(big.Int).SetString(totalStr, 10); parsed && latestSkaVoters > 0 {
-						perVote := new(big.Int).Div(total, big.NewInt(latestSkaVoters))
+			var blockHeight int64
+
+			// Find the latest block specifically for this coin type
+			if len(blockData.ExtraInfo.SSFeeTotalsByCoin) > 0 {
+				if totalStr, ok := blockData.ExtraInfo.SSFeeTotalsByCoin[ct]; ok {
+					if total, parsed := new(big.Int).SetString(totalStr, 10); parsed && blockData.Header.Voters > 0 {
+						perVote := new(big.Int).Div(total, big.NewInt(int64(blockData.Header.Voters)))
 						perBlock = txhelpers.FormatSKAAtoms(perVote)
+						blockHeight = int64(blockData.Header.Height)
 					}
 				}
 			}
+
+			if perBlock == "" {
+				// Fallback: Search backwards through historical summaries for this coin
+				for i := len(sum30) - 1; i >= 0; i-- {
+					if totalStr, ok := sum30[i].SSFeeTotalsByCoin[ct]; ok {
+						if total, parsed := new(big.Int).SetString(totalStr, 10); parsed {
+							if bInfo := exp.dataSource.GetExplorerBlock(ctx, sum30[i].Hash); bInfo != nil && bInfo.BlockBasic.Voters > 0 {
+								perVote := new(big.Int).Div(total, big.NewInt(int64(bInfo.BlockBasic.Voters)))
+								perBlock = txhelpers.FormatSKAAtoms(perVote)
+								blockHeight = int64(sum30[i].Height)
+								break
+							}
+						}
+					}
+				}
+			}
+
 			rewards = append(rewards, types.SKAVoteReward{
 				CoinType:    ct,
 				Symbol:      fmt.Sprintf("SKA%d", ct),
 				PerBlock:    perBlock,
 				Per30Days:   txhelpers.AvgSSFeeRate(sum30S, ct, exp.ChainParams.TicketsPerBlock),
 				PerYear:     txhelpers.AvgSSFeeRate(sumYearS, ct, exp.ChainParams.TicketsPerBlock),
-				BlockHeight: voteRewardsBlockHeight,
+				BlockHeight: blockHeight,
 			})
 		}
 		sort.Slice(rewards, func(i, j int) bool { return rewards[i].CoinType < rewards[j].CoinType })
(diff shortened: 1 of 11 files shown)


edshav (approved):
**Overall:** Solid, well-motivated fix. The core reward calculation bug is real and the fix is correct. The new per-ticket APY approach is more accurate than a block average. A few things worth addressing before merging.

---

### `cmd/dcrdata/internal/explorer/explorer.go` — APY calculation

**Issue 1 — `totalReward` lookup duplication:** When the block is not the current block, the code re-scans `sum30` by height to find `totalReward`. But `blockHeight` and `blockHash` were already resolved during the `perBlock` calculation above, where the matching `sum30[i]` entry was already in hand. Capturing `totalReward` there would avoid the second scan and make the logic easier to follow.

**Issue 2 — `TicketPrice` string format:** `vd.TicketPrice` is scanned directly from the DB and parsed with `new(big.Float).SetString(...)`. If the DB column stores the value as an integer (atoms) rather than a decimal string, `SetString` will silently fail, `parsedP` will be false, all tickets will be skipped, and `perYear` will silently return `"0.000000000000000000"`. Worth confirming the column type and format match what `big.Float.SetString` expects.

**Issue 3 — `rewardPerTicket` divisor:** `rewardPerTicket` is divided by `exp.ChainParams.TicketsPerBlock` before the loop, but the loop iterates over `voteData` which reflects actual voters. If any votes were missed, `len(voteData) < TicketsPerBlock` and the per-ticket reward will be understated. A comment explaining the intended behavior here (or using `len(voteData)` as the divisor) would clarify the design intent.

---

### `db/dcrpg/internal/stakestmts.go` — SQL query

```sql
SELECT v.ticket_price, v.height, t.block_height
FROM votes v 
JOIN tickets t ON v.ticket_hash = t.tx_hash 
WHERE v.block_hash = $1;
```

The PR description says the join uses an "indexed join on ticket_hash" — but the index would need to be on `tickets.tx_hash` (for the join) and ideally on `votes.block_hash` (for the WHERE). Worth verifying both exist with `EXPLAIN ANALYZE` on a populated DB, especially since this runs on every new block.

---

### `pubsub/pubsubhub.go` — PerYear divergence

`pubsubhub` still uses `txhelpers.AvgSSFeeRate(sumYear, ...)` for `PerYear` — the old year-average approach — while `explorer.go` now uses the new per-ticket APY. This means the initial page load and the WebSocket push will show **different `PerYear` values** for the same block. The `Per30Days` removal is consistent, but `PerYear` is now computed two different ways across the two code paths.

---

### `db/dbtypes/types.go` — `VoteTicketData`

Minor: the struct fields have no JSON tags. Not a bug today, but worth being intentional about if this type ever gets serialized.

edshav (approved):
The three issues from the previous review are all addressed. A few new observations from the follow-up commits.

---

### `TicketPrice` float64 precision

The integer-to-decimal conversion uses `float64` as an intermediate:
```go
priceFloat, err := strconv.ParseFloat(priceStr, 64)
priceStr = fmt.Sprintf("%.8f", priceFloat/1e8)
```
`float64` has ~15-16 significant digits, so large atom values may be rounded before reaching `big.Float`. A safer approach would use `big.Int` arithmetic directly to avoid the precision loss. Also worth double-checking the scale factor — `1e8` is used here but `1e18` is used elsewhere in the same file for SKA atoms. Make sure both are intentional and consistent.

---

### Duplicated APY logic in `pubsubhub.go`

The per-ticket APY calculation block is now a near-verbatim copy of the `explorer.go` logic. Any future fix to the formula needs to be applied in two places. Worth extracting into a shared helper (e.g. in `txhelpers`) to avoid drift.

---

### `GetSummaryRange` called twice per coin type in `pubsubhub.go`

In the fallback path, `psh.sourceBase.GetSummaryRange(ctx, startYear, tip)` is called once to find `blockHash`/`perBlock`, and then potentially again inside the APY block to find `totalReward`. That's a year of block summaries fetched twice per coin type per block. Fetching once before the `coinTypes` loop (as `explorer.go` does with `sum30`) would fix this.

---

### Coin supply broadcast removed

The `VARCoinSupply` / `SKACoinSupply` push was removed from `pubsubhub.Store`. If clients rely on WebSocket updates for supply data, this could be a silent regression. Please confirm the supply data is still broadcast via another path.

---

### Minor: typo in comment

```go
// (e.g. via coinbase) even before they arriveT.
```
Stray capital `T` at the end.

edshav (approved):
All four issues from the previous review are resolved. A couple of small observations.

---

### Duplicated APY logic ✅ Fixed
The APY calculation is now extracted into `txhelpers.CalculateAverageTicketAPY`. Both `explorer.go` and `pubsubhub.go` call the shared helper.

---

### `float64` precision for `TicketPrice` ✅ Fixed
The new implementation uses `big.Int.SetString` for the atoms path and `big.ParseFloat` with 256-bit precision for the decimal path. Both are correct.

One minor note: the atoms path divides by `1e8`, while the reward path in `explorer.go`/`pubsubhub.go` divides by `1e18`. Make sure the ticket price denomination (`1e8`) is consistent with what the DB actually stores. If ticket prices are stored in the same atom unit as SKA rewards (`1e18`), the APY ratio will be off by `1e10`. Worth a concrete verification against the DB schema.

---

### `GetSummaryRange` called twice ✅ Fixed
`pubsubhub.go` now reuses `sumYear` throughout.

---

### Coin supply broadcast ✅ Restored

### Typo ✅ Fixed

---

### `SSFeeSummary` struct — new fields
`Hash` and `Height` were added to `SSFeeSummary`. The `toSummaries` helper in `pubsubhub.go` now populates them. If a similar `toSummaries` helper exists in `explorer.go`, verify it's consistent and that `SSFeeSummary.Hash`/`Height` aren't expected there.

---

### `AvgSSFeeRate` removed
Confirm no remaining callers outside this PR — the linter pass suggests it's clean, but worth a quick check if there are any non-compiled consumers.

---

| Area | Status | Note |
|---|---|---|
| Duplicated APY logic | ✅ Fixed | Shared helper in `txhelpers` |
| `TicketPrice` float64 precision | ✅ Fixed | `big.Int` atoms path, `big.ParseFloat` decimal path |
| Ticket price scale factor (`1e8` vs `1e18`) | ⚠️ Verify | Confirm DB stores ticket price in `1e8` atoms, not `1e18` |
| `GetSummaryRange` double call | ✅ Fixed | |
| Coin supply broadcast | ✅ Restored | |
| Typo | ✅ Fixed | |
| `SSFeeSummary` Hash/Height population | 🔧 Minor | Verify `toSummaries` in `explorer.go` is consistent if used |
| `AvgSSFeeRate` removal | �� Minor | Confirm no remaining callers |

yanchenko-igor:
> Confirm no remaining callers outside this PR — the linter pass suggests it's clean, but worth a quick check if there are any non-compiled consumers.
> Area 	Status 	Note
> Duplicated APY logic 	✅ Fixed 	Shared helper in `txhelpers`
> `TicketPrice` float64 precision 	✅ Fixed 	`big.Int` atoms path, `big.ParseFloat` decimal path
> Ticket price scale factor (`1e8` vs `1e18`) 	⚠️ Verify 	Confirm DB stores ticket price in `1e8` atoms, not `1e18`
> `GetSummaryRange` double call 	✅ Fixed 	
> Coin supply broadcast 	✅ Restored 	
> Typo 	✅ Fixed 	
> `SSFeeSummary` Hash/Height population 	🔧 Minor 	Verify `toSummaries` in `explorer.go` is consistent if used
> `AvgSSFeeRate` removal 	�� Minor 	Confirm no remaining callers

All points addressed:
- Ticket Price Scale: Verified that ticket prices are stored in VAR atoms, confirming that the 1e8 scale factor in CalculateAverageTicketAPY is correct.
- SSFeeSummary Consistency: Verified that the toSummaries helper in pubsubhub.go is consistent with the updated struct; explorer.go does not utilize this helper.
- AvgSSFeeRate: Confirmed complete removal, verified by linter.

PR #214, opened 2026-05-01
feat: replace ASR simulation with empirical Vote VAR reward breakdown and ROI

### LINKED ISSUE #205: [DATA] Vote VAR Reward
Добавить расшифровку под значением:

строка субсидии (доля от общей субсидии блока, делённая на количество проголосовавших билетов)
строка fee (доля комиссии, приходящаяся на один билет)
строка ROI (годовая рентабельность)
Subsidy: 3.20000000
Fee: 0.01256400
ROI: 12.02% per year

yanchenko-igor:
Summary
Replaces the theoretical simulateASR approach with empirical data derived directly from the blockchain for Vote VAR rewards.
Key Changes
- Empirical Reward Breakdown: Replaced the single "predicted" reward with a detailed breakdown:
    - Subsidy: Calculated as the network standard 16 VAR divided by the actual number of voters in the block.
    - Fee: Implemented a stable 30-day average of stake fees per vote to avoid volatility.
- Realistic ROI Calculation: 
    - Moved away from theoretical constants to calculate ROI based on the actual age of tickets that voted in the most recent block (VoteHeight - PurchaseHeight).
    - Added CoinbaseMaturity to the ticket age to accurately reflect the total capital lock-up time before rewards are spendable.
- Infrastructure Updates:
    - Updated BlockDataBasic and corresponding database queries to expose Voters, FreshStake, and Revocations.
    - Refactored txhelpers.BlockSSFeeTotals to correctly calculate VAR fees by subtracting the 16 VAR subsidy from total SSGen reward outputs.
- Cleanup: Removed the simulateASR function and all associated ASR fields from HomeInfo and StatsInfo.
Verification
- Verified that the ROI calculation now correctly accounts for the maturity period, reducing artificially inflated values.
- Confirmed that VAR fee summation correctly handles the difference between total rewards and the PoS subsidy.
- Ensured consistency between the REST explorer and WebSocket hub implementations.

### DIFF
--- a/cmd/dcrdata/views/home_voting.tmpl
+++ b/cmd/dcrdata/views/home_voting.tmpl
@@ -91,7 +91,7 @@
 
                 {{/* Row 3 left: Vote VAR Reward */}}
                 <div class="col-12 mb-3 mb-sm-2 mb-md-3 mb-lg-3">
-                    <div class="fs13 text-secondary">Vote VAR Reward (most recent)</div>
+                    <div class="fs13 text-secondary">Vote VAR Reward (most&nbsp;recent)</div>
                     <div class="mono lh1rem fs14-decimal fs24 pt-1 pb-1 d-flex align-items-baseline">
                         <span data-voting-target="bsubsidyPos">
                             {{template "decimalParts" (float64AsDecimalParts $page.Info.VoteVARReward.PerBlock 8 false 2)}}
@@ -134,7 +134,7 @@
 
                 {{/* Full Width: Vote SKA Fee Reward */}}
                 <div class="col-24 mb-3 mb-sm-2 mb-md-3 mb-lg-3">
-                    <div class="fs13 text-secondary">Vote SKA Fee Reward (most recent)</div>
+                    <div class="fs13 text-secondary">Vote SKA Fee Reward (most&nbsp;recent)</div>
                     <div data-voting-target="skaVoteRewards">
                         {{if $page.Info.SKAVoteRewards}}
                             {{range $page.Info.SKAVoteRewards}}
--- a/cmd/dcrdata/views/home_voting.tmpl
+++ b/cmd/dcrdata/views/home_voting.tmpl
@@ -74,7 +74,7 @@
         </div>
         {{/* Row 3 left: Vote VAR Reward */}}
         <div class="col-12 mb-3 mb-sm-2 mb-md-3 mb-lg-3">
-            <div class="fs13 text-secondary">Vote VAR Reward</div>
+            <div class="fs13 text-secondary">Vote VAR Reward (most recent)</div>
             <div class="mono lh1rem fs14-decimal fs24 pt-1 pb-1 d-flex align-items-baseline">
                 <span data-voting-target="bsubsidyPos">
                     {{template "decimalParts" (float64AsDecimalParts $page.Info.VoteVARReward.PerBlock 8 false)}}
@@ -82,9 +82,12 @@
                  <span class="ps-1 unit lh15rem fs13">VAR/Vote</span>
             </div>
             <div class="fs12 lh1rem text-black-50">
-                <span data-voting-target="ticketReward">{{printf "%.2f" $page.Info.VoteVARReward.Per30Days}}%</span> per 30 days
+                Subsidy: {{template "decimalParts" (float64AsDecimalParts $page.Info.VoteVARReward.Subsidy 8 false)}}
             </div>
-            <div class="fs12 lh1rem text-black-50">{{printf "%.2f" $page.Info.VoteVARReward.PerYear}}% per year</div>
+            <div class="fs12 lh1rem text-black-50">
+                Fee: {{template "decimalParts" (float64AsDecimalParts $page.Info.VoteVARReward.Fee 8 false)}}
+            </div>
+            <div class="fs12 lh1rem text-black-50">ROI: {{printf "%.2f" $page.Info.VoteVARReward.ROI}}% per year</div>
         </div>
         {{/* Row 3 right: Total Staked VAR (moved from bottom-left, now below Next Ticket Price Change) */}}
         <div class="col-12 mb-3 mb-sm-2 mb-md-3 mb-lg-3">
--- a/REWARDS_LOGIC.md
+++ b/REWARDS_LOGIC.md
@@ -105,6 +105,6 @@ Verified using `dlv` breakpoints in `txhelpers.RewardsAtBlock` and `explorer.go`
 ### 3. Vote VAR Reward ROI Calculation
 Verified the ROI calculation logic for the most recent block:
 - For each ticket that voted in the block, ROI is calculated as:
-  `ROI = (RewardPerVote / TicketPrice) * (BlocksPerYear / (VoteHeight - PurchaseHeight)) * 100`
+  `ROI = (RewardPerVote / TicketPrice) * (BlocksPerYear / (VoteHeight - PurchaseHeight + CoinbaseMaturity)) * 100`
 - The final displayed ROI is the average of these individual ROIs across all voters in the block.
 
--- a/REWARDS_LOGIC.md
+++ b/REWARDS_LOGIC.md
@@ -54,10 +54,12 @@ The SKA fees collected in a block are distributed among the PoS voters. Miners a
 - **VAR Reward Reduction**: A progress bar showing the current block's position within the subsidy reduction window (`IdxInRewardWindow` / `RewardWindowSize`).
 
 ### Voting Section
-- **Vote VAR Reward**:
-    - **Per Block**: The absolute VAR reward per vote for the next block (`NextBlockSubsidy.PoS / TicketsPerBlock`).
-    - **Per 30 Days**: A linear projection of the current per-vote reward expressed as a profitability percentage.
-    - **Per Year (ASR)**: The Annual Staking Rate computed via `simulateASR`, which runs a simulation of buying and reinvesting tickets over 365 days (accounting for `TicketMaturity` and `CoinbaseMaturity` delays).
+- **Vote VAR Reward (most recent)**: Displays the total reward per vote for the most recent block, broken down by Subsidy and Fee.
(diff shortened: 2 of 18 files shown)


cmd/dcrdata/views/home_voting.tmpl
```diff
@@ -91,7 +91,7 @@
 
                 {{/* Row 3 left: Vote VAR Reward */}}
                 <div class="col-12 mb-3 mb-sm-2 mb-md-3 mb-lg-3">
-                    <div class="fs13 text-secondary">Vote VAR Reward (most&nbsp;recent)</div>
+                    <div class="fs13 text-secondary">Vote VAR Reward (most recent)</div>
```
edshav: `&nbsp;` - is a non breakable space. We use it to keep `(most&nbsp;recent)` on the single text line in case the text wrapped. You shouldn't remove the `&nbsp;`

cmd/dcrdata/views/home_voting.tmpl
```diff
@@ -134,7 +134,7 @@
 
                 {{/* Full Width: Vote SKA Fee Reward */}}
                 <div class="col-24 mb-3 mb-sm-2 mb-md-3 mb-lg-3">
-                    <div class="fs13 text-secondary">Vote SKA Fee Reward (most&nbsp;recent)</div>
+                    <div class="fs13 text-secondary">Vote SKA Fee Reward (most recent)</div>
```
edshav: `&nbsp;` - is a non breakable space. We use it to keep `(most&nbsp;recent)` on the single text line in case the text wrapped. You shouldn't remove the `&nbsp;`

edshav (commented):
Remaining issues to address before merge:

**🟡 Duplicate logic across pipelines (C2 risk)**
The ~70-line ROI computation block is copy-pasted verbatim between \`explorer.go\` and \`pubsubhub.go\`. If one is patched the other will silently drift. Extract into a shared helper (e.g. \`txhelpers.ComputeVoteVARReward\`) so both pipelines stay in sync and the logic is independently testable.

**🟡 VAR fee atom string should use \`ParseInt\`, not \`ParseFloat\`**
In both \`explorer.go\` and \`pubsubhub.go\`, the 30-day fee accumulation parses the atom string with \`strconv.ParseFloat\`. The stored value is an \`int64\`-derived decimal string; use \`strconv.ParseInt\` to stay consistent with the C1 type contract for VAR and avoid any future confusion if this path is reused near SKA values.

**🟡 \`blocksPerYear\` type change — add a comment**
The \`time.Duration\` integer division is now cast to \`float64\` after the fact. The truncation behaviour is preserved, but it's non-obvious. A one-line comment would prevent future readers from second-guessing it.

**🟡 Remove debug log lines before merge**
\`log.Debugf(\"ROI Debug ...\")\` calls are in the hot block-processing path in both \`explorer.go\` and \`pubsubhub.go\`. These will spam production logs on every block.

**🟡 \`TicketReward\` field consistency**
\`HomeInfo.TicketReward\` is still being set in the pubsub path but the explorer path no longer sets it the same way. Verify whether it is still consumed by any template or JS target; if not, remove it from \`HomeInfo\` for consistency with the \`ASR\` removal.

**✅ Positives**
- Removing \`simulateASR\` is a clean win.
- 30-day fee averaging for stability is a good design choice.
- \`CoinbaseMaturity\` inclusion in lock-up age is correct.
- \`ct == 0\` skip in the SKA rewards loop prevents VAR leaking into the SKA section.
- \`ssfee.go\` VAR fee calculation (SSGen outputs minus 16 VAR subsidy) is semantically correct.
- Template and test updates are consistent with the struct changes."

edshav (commented):
## Review (updated after new commits)

**✅ C3 violation — fixed** (voting_controller.js updated)

**✅ Duplicate logic (C2) — addressed**
`ComputeVoteVARReward` extracted into `txhelpers/ssfee.go`. Both `explorer.go` and `pubsubhub.go` now call it.

**✅ `ParseFloat` → `ParseInt` for VAR atoms — fixed**

**✅ `TicketReward` and `ASR` — removed from `HomeInfo` and `StatsInfo`**

**✅ `&nbsp;` — restored** in both "most recent" labels.

**✅ Debug log lines — removed** from the hot path.

---

Remaining issues:

**🟡 `ctx any` parameter is dead weight**
`ComputeVoteVARReward(ctx any, ...)` accepts a context it never uses. Either use `context.Context` and pass it to blocking calls inside, or drop the parameter entirely. `any` as a context type bypasses the type system and signals the parameter was added speculatively.

**🟡 `blockHash string` parameter is also unused**
The `blockHash` argument to `ComputeVoteVARReward` is accepted but never referenced inside the function body. Remove it.

**🟡 `SSFeeSummary` type is now an orphan**
`BlockSummary` was introduced as a replacement and `SSFeeCoinTypes` was updated to accept `[]BlockSummary`, but `SSFeeSummary` still exists in `ssfee.go`. Remove it to avoid confusion.

edshav (commented):
✅ All three remaining issues addressed in the latest commit:
- `ctx any` and `blockHash` parameters removed from `ComputeVoteVARReward`
- `SSFeeSummary` type removed; `BlockSummary` used consistently throughout

PR looks good to merge.

edshav (approved):

edshav (commented):
## Review — new commits (fix: latest block reward, exclude returned ticket values, preserve negative fees)

**✅ Direction is correct** — switching from a 30-day average to the live block's own transactions for the VAR fee is the right call. The `CurrentBlockSubsidy` addition to `BlockExplorerExtraInfo` is clean and the `ParseInt` path is preserved.

---

**🔴 `ComputeTxFeeData` uses raw opcode byte (`0xbb`) for SSGen detection instead of `stake.DetermineTxType`**

```go
if len(out.PkScript) > 0 && out.PkScript[0] == 0xbb {
    isSSGen = true
}
```

Every other place in this codebase — `blockdata.go`, `txhelpers.go`, `pgblockchain.go`, and the existing `ssfee.go` SSFee loop — uses `stake.DetermineTxType(tx)` for this. The raw byte check is fragile: it will misfire on any non-SSGen transaction that happens to start with `0xbb`, and it bypasses the consensus library's authoritative type detection. Replace with:

```go
func ComputeTxFeeData(msgBlock *wire.MsgBlock) []TxFeeData {
    var result []TxFeeData
    for _, tx := range msgBlock.STransactions {
        if stake.DetermineTxType(tx) != stake.TxTypeSSGen {
            continue
        }
        var varOut, varIn int64
        for _, out := range tx.TxOut {
            if out.CoinType == cointype.CoinTypeVAR {
                varOut += out.Value
            }
        }
        for _, vin := range tx.TxIn {
            varIn += vin.ValueIn
        }
        result = append(result, TxFeeData{
            TotalVAROutputs: varOut,
            TotalInputs:     varIn,
            IsSSGen:         true,
        })
    }
    return result
}
```

Note: SSGen transactions only appear in `STransactions`, not `Transactions`, so the `append(msgBlock.Transactions, msgBlock.STransactions...)` in the current implementation is also unnecessary.

---

**🟡 `SumWinningTicketPrices` result is immediately discarded in `pubsubhub.go`**

```go
winningTicketPricesSum := txhelpers.SumWinningTicketPrices(txVoteData, blockData.WinningTickets)
_ = winningTicketPricesSum
```

This is dead code — the function is called, the result thrown away. Either wire it into the reward calculation or remove the call. As-is it just burns CPU on every block.

---

**�� Absolute value of negative fee is semantically questionable**

```go
if actualTotalFees < 0 {
    actualTotalFees = -actualTotalFees
}
```

A negative net redistribution means ticket holders paid more in inputs than they received in outputs (consolidation cost). Displaying `abs(fee)` as a positive reward overstates what voters actually earned. The correct behaviour for a negative net is to show `feePerVote = 0` (no fee reward this block), not the absolute value. The test `TestComputeVoteVARReward_NegativeFee` explicitly asserts `result.Fee > 0` for a negative input — this test is asserting the wrong thing.

---

**🟡 `dbCtypes` typo in comment**

`db/dbtypes/conversion.go` line ~72: `// Create the dbCtypes.Block structure` — should be `dbtypes`.

edshav (commented):
## Review — new commits (use stake.DetermineTxType, cap negative fees, include SSRtx/type=7)

**✅ `stake.DetermineTxType` — fixed.** Raw `0xbb` opcode check replaced with the proper library call.

**✅ Dead code removed.** `SumWinningTicketPrices` / `_ = winningTicketPricesSum` gone from `pubsubhub.go`.

**✅ Negative fee capped to 0** with an error log instead of taking the absolute value. Semantics are now correct.

**✅ Typo fixed.** `dbCtypes` → `dbtypes` in `conversion.go`.

**✅ SSRtx and type=7 transactions included** in the fee calculation, with real-block test coverage.

---

One issue remaining:

**🟡 Magic number `7` should use the named constant**

In `BlockSSFeeTotals` and `ComputeTxFeeData`, type=7 is hardcoded as a literal:

```go
isMisclassifiedSSRtx := txType == 7 // SSFee type, but actually SSRtx
```

`TxTypeSSFee` is indeed 7 in the iota (Regular=0, SStx=1, SSGen=2, SSRtx=3, TAdd=4, TSpend=5, TreasuryBase=6, SSFee=7), so the value is correct. But the comment calling it "misclassified SSRtx" is misleading — these are genuine `TxTypeSSFee` transactions (SKA stake fee distributions). Use the named constant and update the comment:

```go
// TxTypeSSFee: SKA stake fee distribution transactions
isTxSSFee := txType == int(stake.TxTypeSSFee)
```

Same in the test fixtures: `TxType: 7` → `TxType: int(stake.TxTypeSSFee)`.

---

Otherwise this is clean. Happy to approve once that literal is replaced.

edshav (commented):
✅ Both issues resolved:
- Magic `7` replaced with `int(stake.TxTypeSSFee)` throughout code and tests.
- "Misclassified SSRtx" comment corrected to "SKA stake fee distribution", with TODO markers to track the upstream node fix.

No further issues. This PR is good to merge.

PR #225, opened 2026-05-06
feat: provide multi-coin data for block page template and verbose API

yanchenko-igor:
- Add CoinAmount, CoinCount types and TotalSentByCoin, RegularCoinCounts, FeesByCoin fields to BlockInfo
- Add CoinType, TicketStatus fields to TrimmedTxInfo for stake fees and revocation rows
- Add helper functions for ordered slice generation (VAR first, SKA sorted, zero omitted)
- Update GetExplorerBlock to populate new template-facing fields from block summary
- Extend /api/block/{height}/verbose with total_sent and fees map objects
- Update block template: TotalSent iterates coins, Fees shows coin type, Revocations add Status column, StakeFees add coin type, Transactions remove Mixed column
- Add unit tests for ordered slice generation and API tests for verbose endpoint

### DIFF
--- a/cmd/dcrdata/internal/api/apiroutes.go
+++ b/cmd/dcrdata/internal/api/apiroutes.go
@@ -585,31 +585,30 @@ func (c *appContext) getBlockVerbose(w http.ResponseWriter, r *http.Request) {
 	// Get summary for multi-coin data
 	summary := c.DataSource.GetSummaryByHash(ctx, hash, false)
 
-	// Build extended response with total_sent and fees maps
-	response := make(map[string]interface{})
+	// Build response struct with embedded verbose result
+	type verboseResponse struct {
+		*chainjson.GetBlockVerboseResult
+		TotalSent map[string]string `json:"total_sent,omitempty"`
+		Fees      map[string]string `json:"fees,omitempty"`
+	}
 
-	// Marshal the verbose result into a map
-	if b, err := json.Marshal(blockVerbose); err == nil {
-		json.Unmarshal(b, &response)
+	response := verboseResponse{
+		GetBlockVerboseResult: blockVerbose,
 	}
 
 	// Add total_sent map (coin type key -> amount string)
 	if summary != nil && summary.CoinAmounts != nil {
-		totalSent := make(map[string]string)
+		response.TotalSent = make(map[string]string)
 		for ct, amt := range summary.CoinAmounts {
 			if amt != "0" && amt != "" {
-				totalSent[fmt.Sprintf("%d", ct)] = amt
+				response.TotalSent[fmt.Sprintf("%d", ct)] = amt
 			}
-		}
-		if len(totalSent) > 0 {
-			response["total_sent"] = totalSent
 		}
 
 		// Add fees map (coin type key -> fee string)
 		if summary.MiningFee != nil && *summary.MiningFee != 0 {
-			fees := make(map[string]string)
-			fees["0"] = fmt.Sprintf("%d", *summary.MiningFee)
-			response["fees"] = fees
+			response.Fees = make(map[string]string)
+			response.Fees["0"] = fmt.Sprintf("%d", *summary.MiningFee)
 		}
 	}
 
--- a/cmd/dcrdata/internal/api/apiroutes.go
+++ b/cmd/dcrdata/internal/api/apiroutes.go
@@ -606,6 +606,7 @@ func (c *appContext) getBlockVerbose(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Add fees map (coin type key -> fee string)
+		// VAR fees from MiningFee (total VAR tx fees in block); SKA fees from SSFeeTotalsByCoin
 		response.Fees = make(map[string]string)
 
 		// VAR fees from MiningFee
--- a/cmd/dcrdata/internal/api/apiroutes.go
+++ b/cmd/dcrdata/internal/api/apiroutes.go
@@ -582,7 +582,38 @@ func (c *appContext) getBlockVerbose(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	writeJSON(w, blockVerbose, m.GetIndentCtx(r))
+	// Get summary for multi-coin data
+	summary := c.DataSource.GetSummaryByHash(ctx, hash, false)
+
+	// Build extended response with total_sent and fees maps
+	response := make(map[string]interface{})
+
+	// Marshal the verbose result into a map
+	if b, err := json.Marshal(blockVerbose); err == nil {
+		json.Unmarshal(b, &response)
+	}
+
+	// Add total_sent map (coin type key -> amount string)
+	if summary != nil && summary.CoinAmounts != nil {
+		totalSent := make(map[string]string)
+		for ct, amt := range summary.CoinAmounts {
+			if amt != "0" && amt != "" {
+				totalSent[fmt.Sprintf("%d", ct)] = amt
+			}
+		}
+		if len(totalSent) > 0 {
+			response["total_sent"] = totalSent
+		}
+
+		// Add fees map (coin type key -> fee string)
+		if summary.MiningFee != nil && *summary.MiningFee != 0 {
+			fees := make(map[string]string)
+			fees["0"] = fmt.Sprintf("%d", *summary.MiningFee)
+			response["fees"] = fees
+		}
+	}
+
+	writeJSON(w, response, m.GetIndentCtx(r))
 }
 
 func (c *appContext) getVoteInfo(w http.ResponseWriter, r *http.Request) {
--- a/cmd/dcrdata/internal/api/apiroutes.go
+++ b/cmd/dcrdata/internal/api/apiroutes.go
@@ -606,9 +606,25 @@ func (c *appContext) getBlockVerbose(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Add fees map (coin type key -> fee string)
+		response.Fees = make(map[string]string)
+
+		// VAR fees from MiningFee
 		if summary.MiningFee != nil && *summary.MiningFee != 0 {
-			response.Fees = make(map[string]string)
 			response.Fees["0"] = fmt.Sprintf("%d", *summary.MiningFee)
+		}
+
+		// SKA fees from SSFeeTotalsByCoin (stake fee distribution)
+		if summary.SSFeeTotalsByCoin != nil {
+			for ct, fee := range summary.SSFeeTotalsByCoin {
+				if fee != "0" && fee != "" {
+					response.Fees[fmt.Sprintf("%d", ct)] = fee
+				}
+			}
+		}
+
+		// Remove empty fees map
+		if len(response.Fees) == 0 {
+			response.Fees = nil
 		}
 	}
 
--- a/cmd/dcrdata/internal/api/apiroutes_test.go
+++ b/cmd/dcrdata/internal/api/apiroutes_test.go
@@ -261,6 +261,71 @@ func TestBlockVerbose_FeesMap(t *testing.T) {
 	}
 }
 
+// blockVerboseSKAFeeDS provides SSFeeTotalsByCoin for SKA fee testing.
+type blockVerboseSKAFeeDS struct {
+	blockVerboseDS
+}
+
+func (blockVerboseSKAFeeDS) GetSummaryByHash(_ context.Context, hash string, _ bool) *apitypes.BlockDataBasic {
+	fee := int64(12345678)
+	return &apitypes.BlockDataBasic{
+		Height: 100,
+		Hash:   hash,
+		CoinAmounts: map[uint8]string{
+			0: "100000000",
+			1: "1000000000000000000",
+		},
+		MiningFee: &fee,
+		SSFeeTotalsByCoin: map[uint8]string{
+			1: "500000000000000000", // SKA1: 0.5 SKA in atoms
+			2: "300000000000000000", // SKA2: 0.3 SKA in atoms
+		},
+	}
+}
+
+func TestBlockVerbose_SKAFeesFromSSFeeTotals(t *testing.T) {
+	app := &appContext{DataSource: blockVerboseSKAFeeDS{}}
+	mux := NewAPIRouter(app, "", false, false)
+
+	req := httptest.NewRequest(http.MethodGet, "/block/100/verbose", nil)
+	w := httptest.NewRecorder()
+	mux.ServeHTTP(w, req)
+
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d: %s", w.Code, w.Body.String())
+	}
+
+	var result map[string]interface{}
+	if err := json.NewDecoder(w.Body).Decode(&result); err != nil {
+		t.Fatalf("decode: %v", err)
+	}
+
+	fees, ok := result["fees"].(map[string]interface{})
+	if !ok {
+		t.Fatal("fees field missing or not a map")
+	}
+
+	// VAR fees should be present
+	if fees["0"] == nil {
+		t.Error("fees missing VAR (key 0)")
+	}
+
+	// SKA fees from SSFeeTotalsByCoin should be present
+	if fees["1"] == nil {
+		t.Error("fees missing SKA1 (key 1) from SSFeeTotalsByCoin")
+	}
+	if fees["1"] != "500000000000000000" {
+		t.Errorf("fees[1] = %v, want 500000000000000000", fees["1"])
+	}
+
+	if fees["2"] == nil {
+		t.Error("fees missing SKA2 (key 2) from SSFeeTotalsByCoin")
+	}
+	if fees["2"] != "300000000000000000" {
+		t.Errorf("fees[2] = %v, want 300000000000000000", fees["2"])
+	}
+}
+
 // blockVerboseZeroSKADS provides zero-value SKA for omission test.
 type blockVerboseZeroSKADS struct {
 	blockVerboseDS
--- a/cmd/dcrdata/internal/api/apiroutes_test.go
+++ b/cmd/dcrdata/internal/api/apiroutes_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	apitypes "github.com/monetarium/monetarium-explorer/api/types"
+	chainjson "github.com/monetarium/monetarium-node/rpc/jsonrpc/types"
 )
 
 // blockSummaryDS overrides GetSummaryByHash to return multi-coin data.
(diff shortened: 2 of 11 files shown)


edshav (commented):
## Review of #225 against requirements in #223

Thanks for the work. Below is a combined review of code quality plus a gap analysis against the acceptance criteria in #223. Roughly half of the issue's requirements are met; the rest are either missing populators, contradict the precision/ordering rules, or were not removed as required.

### Requirements coverage vs #223

**Met**
- `TotalSent` ordered slice with VAR-first / sorted-SKA / zero-omitted ordering — `TotalSentByCoinFromMap`, populated at `db/dcrpg/pgblockchain.go:6418`.
- `RegularCoinCounts` ordered slice with the same rules — `RegularCoinCountsFromCoinRows`.
- Verbose API `total_sent` map with zero-SKA omission — `cmd/dcrdata/internal/api/apiroutes.go:597-606`.

**Partially met / behaves wrong**
- **`Fees` slice does not satisfy the rules.** `FeesByCoinFromMiningFee` always returns a single hardcoded `{"VAR", atoms}` element and **does not omit a zero value** (a coinbase-only block produces `[{"VAR","0"}]`). It also cannot represent SKA fees because the input is a single VAR `float64`. Per `CLAUDE.md`, *"transaction fees are paid in the same coin as the transaction itself; don't write code that assumes fees are always VAR"* — the helper is shaped against this invariant.
- **Precision constraint.** The issue says *"no formatting, no decimal insertion, no float conversion before the template boundary."* `FeesByCoinFromMiningFee(miningFee float64)` then does `int64(miningFee * 1e8)`. VAR fits safely today, but the helper's signature is the wrong shape for the rule, and it's the entry point that will need to take SKA atoms next.
- **Verbose API `fees` map** is single-key (`"0"`) only — same root cause as above.

**Not implemented**
- **Revocation `Status`.** `TrimmedTxInfo.TicketStatus` is added but never populated. `trimmedTxInfoFromMsgTx` (`pgblockchain.go:6328`) doesn't set it; the `stake.TxTypeSSRtx` branch (`pgblockchain.go:6474`) appends unchanged. The data is already available via `retrieveTicketStatusByHash` (`pgblockchain.go:1525`, used by the ticket page).
- **Stake-fee rows: coin type on `Total`/`Fee`.** `TrimmedTxInfo.CoinType` added but never set for `stake.TxTypeSSFee`.
- **Transactions rows: coin type on `Total`.** Not populated in the regular-tx loop (`pgblockchain.go:6485-6502`). It's trivially derivable from `msgTx.TxOut[0].CoinType` (the same pattern exists at `pgblockchain.go:5252`).
- **`Mixed` field removed from context.** Still computed (`pgblockchain.go:6483, 6501, 6510`), still on `BlockInfo.TotalMixed` (`explorertypes.go:711`), and `MixedPercent` is still set in the home view-model.
- **No integration test** that builds a `BlockInfo` for a mixed VAR+SKA block and asserts the new fields. Unit tests cover helpers in isolation; verbose-API tests use a hand-crafted DS mock and don't exercise `GetExplorerBlock` wiring.

### Code-quality issues (independent of the gap above)

- **`getBlockVerbose` marshals into `map[string]interface{}`** (`apiroutes.go:589-594`). Re-encoding through a generic map decodes every numeric field through `float64`, silently losing precision on `int64` fields in the verbose result. Replace with an embedded-struct wrapper:
  ```go
  resp := struct {
      *chainjson.GetBlockVerboseResult
      TotalSent map[string]string `json:"total_sent,omitempty"`
      Fees      map[string]string `json:"fees,omitempty"`
  }{GetBlockVerboseResult: blockVerbose}
  ```
  Also, the `json.Unmarshal` error at line 593 is silently swallowed.
- **`RegularCoinCountsFromCoinRows` round-trips ints through strings** (`explorertypes.go:258-288`): builds `[]CoinAmount` (string `Amount`) and then `Sscanf`s each back to `int` to fill `[]CoinCount`. Build `[]CoinCount` directly. The `Sscanf` error is also discarded.
- **`FeesByCoinFromMiningFee` signature** should take `int64` atoms, not `float64` DCR — atoms originate as `int64` upstream, the float multiplication is needlessly lossy.
- **`CoinTypeFromSymbol` uses `fmt.Sscanf`** (`explorertypes.go:188-194`). Prefer `strconv.Atoi(symbol[3:])` — stricter parsing, faster, no partial-match surprises.
- **New `CoinAmount` / `CoinCount` types deviate from the codebase's established CoinType convention.** Every existing struct with a `coin_type` JSON field uses `CoinType uint8`:
  - `api/types/apitypes.go:111,126,156,235,397,447`
  - `db/dbtypes/types.go:1237 // 0=VAR, 1-255=SKA`, 1282, 2056
  - `explorer/types/explorertypes.go:618` (`CoinRowData`), `:641` (`SKACoinSupplyEntry`), `:1428`

  Every coin-keyed map is `map[uint8]…` (`CoinAmounts`, `CoinTxStats`, `CoinStats`, `SSFeeTotalsByCoin`, `MempoolCoinStats`) — which is exactly what the verbose API in this PR emits as `"0"`/`"1"` keys, so the verbose path is consistent. The outliers are the new `CoinAmount.CoinType string` and `CoinCount.CoinType string`. The established pattern, visible in `CoinRowData`, is `CoinType uint8` as the canonical identifier plus an optional `Symbol string` for display. Recommend changing both new structs to `CoinType uint8`, adding a `Symbol string` field if the template needs the label (or letting templates call `CoinTypeSymbol(.CoinType)` for display).
- **Test package change** `package types_test` → `package types` (`blockinfo_test.go:1`) doesn't appear necessary; the new tests don't access unexported names. Keeping black-box parity with the rest of the suite would be cleaner.
- **`TrimmedTxInfo.CoinType` / `TicketStatus`** added without a populator are latent fields — link them to a follow-up PR or fill them in this one.

### Suggested actions before merge

1. Populate `TrimmedTxInfo.CoinType` (regular txs, stake fees) and `TicketStatus` (revocations via `retrieveTicketStatusByHash`).
2. Remove `Mixed`/`TotalMixed`/`MixedPercent` from the block context as #223 requires.
3. Reshape `FeesByCoinFromMiningFee` to accept per-coin atoms (or a map), apply zero-omission and ordering, and take `int64` rather than `float64`. Update the verbose API path to match.
4. Replace the marshal-into-map in `getBlockVerbose` with the embedded-struct wrapper above.
5. Simplify `RegularCoinCountsFromCoinRows` to build `[]CoinCount` directly.
6. Change `CoinAmount` / `CoinCount` to use `CoinType uint8` (with optional `Symbol string`) to align with the rest of the codebase.
7. Add an integration test for `GetExplorerBlock` against a mixed VAR+SKA block fixture, asserting `TotalSentByCoin` / `RegularCoinCounts` / `FeesByCoin` / `Revs[i].TicketStatus` are populated correctly.

I don't think this PR currently satisfies the acceptance criteria of #223. Once the populators land and `Fees`/`Mixed` are addressed, it should be close.

edshav (commented):
## Re-review of new commits (`9b1b990d` through `71544a80`)

Most prior feedback is addressed cleanly. One critical bug in the new SKA-fee aggregation, plus a few smaller items.

### Addressed well

- **CoinType convention** (`9b1b990d`): `CoinAmount`/`CoinCount` now use `CoinType uint8` plus optional `Symbol string`, matching the rest of the codebase. ✓
- **Verbose API precision** (`9b1b990d`): replaced `map[string]interface{}` round-trip with an embedded-struct wrapper. ✓
- **`RegularCoinCountsFromCoinRows`** (`9b1b990d`): now builds `[]CoinCount` directly, no `Sscanf` round-trip. ✓
- **`CoinTypeFromSymbol`** (`9b1b990d`): switched to `strconv.Atoi`. ✓
- **`TrimmedTxInfo.CoinType` populators** (`9b1b990d`): regular txs and stake fees now set `CoinType` from `msgTx.TxOut[0].CoinType`. ✓
- **`TicketStatus` populator** (`a2656249`): revocations resolve via `retrieveTicketStatusByHash` against the prevout hash — same source the ticket page uses. ✓
- **`Mixed` computation removed** (`9b1b990d`): `totalMixed`/`block.TotalMixed` no longer computed. ✓ (see nit below)
- **Multi-coin fee shape** (`71544a80`): replaced `FeesByCoinFromMiningFee(float64)` with `FeesByCoinFromAmounts(map[uint8]string)` — correct shape, applies zero-omission and ordering. ✓

### Critical bug — SKA fee aggregation captures nothing

In `db/dcrpg/pgblockchain.go` (commit `71544a80`):

```go
for _, tx := range block.Tx {
    if tx.CoinType != 0 && tx.FeeRaw != "" && tx.FeeRaw != "0" {
        feesMap[tx.CoinType] = tx.FeeRaw   // (a) reads FeeRaw, (b) overwrites
    }
}
for _, tx := range block.StakeFees {
    if tx.CoinType != 0 && tx.FeeRaw != "" && tx.FeeRaw != "0" {
        feesMap[tx.CoinType] = tx.FeeRaw
    }
}
```

Two problems:

1. **`tx.FeeRaw` is always `"0"` for SKA at this point.** `makeExplorerTxBasic` hardcodes it, with a comment that says exactly why (`pgblockchain.go:6276-6278`):
   ```go
   // SKA Fee calculation skipped here; handled in GetExplorerTx and GetAPITransaction
   // which have access to the DB client for prevout lookups.
   tx.FeeRaw = "0"
   ```
   The `tx.FeeRaw != "0"` guard is therefore always false for SKA, and the `block.FeesByCoin` slice will never contain SKA entries via the template path. The commit's stated goal is silently unmet.

2. **Even if `FeeRaw` were populated, the aggregation overwrites instead of sums.** `feesMap[tx.CoinType] = tx.FeeRaw` replaces on each iteration. With N SKA1 transactions, only the last tx's fee survives. SKA atoms are 18-decimal, exceeding `int64`, so the sum needs `big.Int` arithmetic.

To actually compute per-coin SKA fees in `GetExplorerBlock`, you need either:
- A path that does the prevout lookup the way `GetExplorerTx` does (around `pgblockchain.go:6755`, the `totalInSKA - totalOutSKA` calc), batched over all block txs; or
- A summary field computed at collection time (analogous to `MiningFee`) that the explorer can read directly — preferable for performance.

### Inconsistent fee semantics between template and API

- **Template path** (broken, see above): aggregates per-tx `FeeRaw`.
- **API path** (`apiroutes.go`, commit `71544a80`): uses `summary.SSFeeTotalsByCoin`, which per its docstring is *"total SKA atoms distributed via TxTypeSSFee per coin type"* — i.e. SSFee distribution amounts, not per-tx SKA fees.

The two paths compute conceptually different quantities for the same `fees` field. Pin down which one is correct (likely "total SKA fees paid by SKA txs in the block"), then make both paths use that single source.

### Smaller issues

- **TicketStatus does N+1 DB queries** (`a2656249`, `pgblockchain.go:6474-6486`). Each revocation triggers `retrieveTicketStatusByHash`. Revocations per block are bounded (low single digits), so likely fine, but worth a comment or a batched lookup if this ever shows up in profiles.
- **`retrieveTicketStatusByHash` signature change** (`b257610e`): moving the unused `id uint64` to a local var silences `unparam` but the row scan still pulls the column. If the column isn't actually needed by callers, the SQL itself (`SelectTicketStatusByHash`) could drop it; otherwise the warning was right and the fix is purely cosmetic.
- **`BlockInfo.TotalMixed int64` still on the struct** (`explorertypes.go:581`); only the computation was removed. `MixedPercent` (`:624`) and the explorer's `MixedPercent = ...` assignment (`explorer.go:934`) also remain. If "Mixed field removed from context" in #223 means the struct field too, those need to go; if it means template-only, this is fine — please confirm scope.
- **No tests for the new fee-aggregation logic** — a unit test asserting `block.FeesByCoin` for a mixed VAR+SKA fixture would have caught issue #1 above. Same for the `tx.CoinType` populators on regular/stake-fee txs.

### Overall

Strong cleanup pass on the API/types side. The SKA-fee aggregation needs to be reworked against a real fee source (or fixed at the `summary` level) — as it stands, the "SKA fees per coin type" feature is wired up but doesn't deliver SKA fees. Once that and the template/API consistency are settled, this should satisfy #223 modulo the `Mixed`-field-on-struct interpretation question.

edshav (commented):
## Re-review of new commits (`9b1b990d` through `71544a80`)

Most prior feedback is addressed cleanly. One critical bug in the new SKA-fee aggregation, plus a few smaller items. Suggested fixes inline.

### Addressed well

- **CoinType convention** (`9b1b990d`): `CoinAmount`/`CoinCount` now use `CoinType uint8` plus optional `Symbol string`, matching the rest of the codebase. ✓
- **Verbose API precision** (`9b1b990d`): replaced `map[string]interface{}` round-trip with an embedded-struct wrapper. ✓
- **`RegularCoinCountsFromCoinRows`** (`9b1b990d`): builds `[]CoinCount` directly, no `Sscanf` round-trip. ✓
- **`CoinTypeFromSymbol`** (`9b1b990d`): switched to `strconv.Atoi`. ✓
- **`TrimmedTxInfo.CoinType` populators** (`9b1b990d`): regular txs and stake fees now set `CoinType` from `msgTx.TxOut[0].CoinType`. ✓
- **`TicketStatus` populator** (`a2656249`): revocations resolve via `retrieveTicketStatusByHash` against the prevout hash — same source the ticket page uses. ✓
- **Mixed computation removed** (`9b1b990d`): `totalMixed`/`block.TotalMixed` no longer computed. ✓ (struct field nit below)
- **Multi-coin fee shape** (`71544a80`): `FeesByCoinFromAmounts(map[uint8]string)` is the correct shape, applies zero-omission and ordering. ✓

### Critical bug — SKA fee aggregation captures nothing

In `db/dcrpg/pgblockchain.go` (commit `71544a80`):

```go
for _, tx := range block.Tx {
    if tx.CoinType != 0 && tx.FeeRaw != "" && tx.FeeRaw != "0" {
        feesMap[tx.CoinType] = tx.FeeRaw   // (a) FeeRaw is "0" for SKA, (b) overwrites
    }
}
```

Two problems:

1. **`tx.FeeRaw` is always `"0"` for SKA at this point.** `makeExplorerTxBasic` hardcodes it (`pgblockchain.go:6276-6278`):
   ```go
   // SKA Fee calculation skipped here; handled in GetExplorerTx and GetAPITransaction
   // which have access to the DB client for prevout lookups.
   tx.FeeRaw = "0"
   ```
   The `tx.FeeRaw != "0"` guard is always false for SKA, so `block.FeesByCoin` will never contain SKA entries via the template path.

2. **Even if `FeeRaw` were populated, the aggregation overwrites instead of sums.** `feesMap[tx.CoinType] = tx.FeeRaw` replaces on each iteration. SKA atoms are 18-decimal, exceeding `int64`, so summation needs `big.Int`.

**Suggested fix — preferred:** compute SKA fees once at collection time and stash on the block summary (analogous to `MiningFee`). Both `GetExplorerBlock` and `getBlockVerbose` then consume the same field.

In whichever package builds `dbtypes.BlockDataBasic`/`apitypes.BlockDataBasic` (likely `blockdata.Collector` or where `coinRowsFromSummary` is fed):

```go
// add to BlockDataBasic alongside CoinAmounts / SSFeeTotalsByCoin:
FeesByCoin map[uint8]string `json:"fees_by_coin,omitempty"` // VAR+SKA fee atoms
```

Populated by aggregating per-tx fees with `big.Int`, using the same `totalIn - totalOut` calc that `GetExplorerTx` already does at `pgblockchain.go:6755`. Then in `GetExplorerBlock`:

```go
block.FeesByCoin = exptypes.FeesByCoinFromAmounts(summary.FeesByCoin)
```

And in `getBlockVerbose`, drop the ad-hoc `MiningFee` + `SSFeeTotalsByCoin` merge and just emit `summary.FeesByCoin` directly.

**Suggested fix — minimal (if collector changes are out of scope):** do the prevout-lookup-and-sum in `GetExplorerBlock` itself, using `big.Int`:

```go
feeSums := make(map[uint8]*big.Int)
// VAR (existing aggregate)
if atoms := int64(block.MiningFee * 1e8); atoms > 0 {
    feeSums[0] = big.NewInt(atoms)
}
// SKA: per-tx in - per-tx out, summed
for _, tx := range append(append([]*exptypes.TrimmedTxInfo{}, block.Tx...), block.StakeFees...) {
    if tx.CoinType == 0 || tx.Coinbase {
        continue
    }
    fee := computeSKAFee(ctx, pgb, tx) // mirrors GetExplorerTx logic at L6755
    if fee.Sign() > 0 {
        if feeSums[tx.CoinType] == nil { feeSums[tx.CoinType] = new(big.Int) }
        feeSums[tx.CoinType].Add(feeSums[tx.CoinType], fee)
    }
}
feesMap := make(map[uint8]string, len(feeSums))
for ct, sum := range feeSums { feesMap[ct] = sum.String() }
block.FeesByCoin = exptypes.FeesByCoinFromAmounts(feesMap)
```

This mirrors the existing per-tx logic but is N+M prevout lookups per block render — acceptable for now; the collector-time approach is the long-term home.

Either way: **add a unit/integration test** covering a block with multiple SKA-typed txs of the same coin type, asserting fees sum (not overwrite). The current bug would have been caught by such a test.

### Inconsistent fee semantics between template and API

- Template path (broken): per-tx `FeeRaw` aggregation.
- API path (`apiroutes.go`, `71544a80`): `summary.SSFeeTotalsByCoin` — per its docstring, *"total SKA atoms distributed via TxTypeSSFee per coin type"*. That's the SSFee **distribution** amount, not the fees paid by SKA txs in the block.

The two paths compute conceptually different quantities for the same `fees` field.

**Suggested fix:** the collector-time `summary.FeesByCoin` field above is the single source. Both paths read it. Drop the ad-hoc merge in `getBlockVerbose`:

```go
// in getBlockVerbose, after summary fetch:
if summary != nil && len(summary.FeesByCoin) > 0 {
    response.Fees = make(map[string]string, len(summary.FeesByCoin))
    for ct, amt := range summary.FeesByCoin {
        if amt != "0" && amt != "" {
            response.Fees[fmt.Sprintf("%d", ct)] = amt
        }
    }
}
```

### Smaller issues

- **TicketStatus N+1 DB queries** (`a2656249`, `pgblockchain.go:6474-6486`). Each revocation hits `retrieveTicketStatusByHash`. Bounded (~5/block), so likely fine.

  *Suggested fix if it ever matters:* add a batched variant:
  ```go
  func retrieveTicketStatusesByHashes(ctx context.Context, db *sql.DB, hashes []dbtypes.ChainHash) (map[dbtypes.ChainHash]dbtypes.TicketPoolStatus, error)
  // SQL: SELECT ticket_hash, pool_status FROM tickets WHERE ticket_hash = ANY($1)
  ```
  Call once before the SSRtx loop. Skip until a profile justifies it.

- **`retrieveTicketStatusByHash` signature change** (`b257610e`): moving unused `id uint64` to a local silences `unparam` cosmetically but the SQL still selects it. If no caller needs the column, drop it from `SelectTicketStatusByHash` too. Otherwise the lint fix is fine as-is.

- **`BlockInfo.TotalMixed int64` still on the struct** (`explorertypes.go:581`); `MixedPercent` (`:624`) and `explorer.go:934` also remain. Computation gone but type still bears the field.

  *Suggested fix:* delete `TotalMixed` from `BlockInfo`, `MixedPercent` from `HomeInfo`, and the assignment at `explorer.go:934`. Mixed-tx tracking is a Decred-specific concept that doesn't apply to Monetarium; partial removal is worse than full removal because it leaves dead state. If "context" in #223 means template-only, please confirm scope explicitly.

- **No tests for new fee-aggregation logic, nor for `tx.CoinType` populators on regular/stake-fee txs.** A fixture-based `GetExplorerBlock` test (block with VAR + multiple SKA1 + one SKA2 txs, asserting `block.FeesByCoin` and per-tx `CoinType`) would close the gap and lock in the fix.

### Overall

Strong cleanup pass on API/types. The SKA-fee path is the remaining blocker — preferred fix is a `summary.FeesByCoin` field at collection time, fallback is per-tx prevout-lookup-and-sum in `GetExplorerBlock`. Either way, sums must use `big.Int`, and a test should exercise the multi-tx-per-coin case.

edshav (commented):
## Re-review of new commits (`a09177fa` through `b984de1f`)

The critical bug from the previous review is resolved. The pivot to `summary.SSFeeTotalsByCoin` for SKA fees in `8917acc4` is the right call — both paths now read from the same field, so the template/API-divergence concern goes away too. Mixed-field cleanup is also properly carried through to the structs and the `ChartsUpdated` assignment.

A couple of follow-ups remain.

### Addressed well
- **SKA fee aggregation** (`8917acc4`): no longer reads `tx.FeeRaw` (which is hardcoded to `"0"` for SKA at this stage). Sourced from `summary.SSFeeTotalsByCoin` instead.
- **Template/API consistency** (`8917acc4`): both paths read `summary.SSFeeTotalsByCoin` for SKA fees — single source of truth.
- **`TotalMixed` removed from `BlockInfo`** (`8917acc4`): struct field gone, computation gone, assignment gone.
- **`MixedPercent` removed from `HomeInfo`** (`8917acc4`): field gone, plus the `ChartsUpdated` assignment in `explorer.go:931`.
- **API test for SKA fees** (`8917acc4`): `TestBlockVerbose_SKAFeesFromSSFeeTotals` asserts SKA1/SKA2 fee values come through with the right keys and values.
- **Aggregation/refactor commits** (`a09177fa`, `2370a66f`, `b984de1f`): the `AggregateFeesFromTxSlice` helper was added and then correctly removed once `8917acc4` made the source change.

### Items still open

#### 1. Stale godoc comment on `FeesByCoinFromAmounts`
`explorer/types/explorertypes.go:282-285` has two godoc paragraphs — the first is leftover from the deleted `FeesByCoinFromMiningFee`:

```go
// FeesByCoinFromMiningFee creates a fees slice from the VAR mining fee.
// For now, only VAR fees are supported; SKA fees may be added later.
// FeesByCoinFromAmounts creates an ordered fee slice from a map of coin type -> fee atoms.
// VAR (0) is first, then SKA types in ascending order, zero-value SKA omitted.
func FeesByCoinFromAmounts(feeAmounts map[uint8]string) []CoinAmount {
```

Drop the first two lines.

#### 2. Semantic — does `SSFeeTotalsByCoin` mean "block fees"?
The field's docstring (`api/types/apitypes.go:770`, `db/dbtypes/types.go:2215`) says:

> "total SKA atoms distributed via TxTypeSSFee per coin type"

That's the SSFee **distribution** amount (output of stake-fee transactions), not the per-tx fees paid by SKA-typed transactions in the block. Both interpretations may be reasonable to surface in the "Fees" card, but they are different quantities. Two questions:

- Is this the value you want under the user-facing "Fees" label? If the intent is "fees paid by SKA txs in this block", this isn't it (SKA per-tx fees compute via the prevout-input vs. output diff, the same way `GetExplorerTx` does at `pgblockchain.go:6755`).
- If `SSFeeTotalsByCoin` *is* the right value, please add a one-line comment in `GetExplorerBlock` and the API handler describing the semantic, so future readers don't conflate it with per-tx fees.

This isn't a blocker for #223's wording (the issue says "Fees ... ordered slice ... per coin type" without nailing down the semantic), but it should be settled explicitly before users start reading values from it.

#### 3. No template-path test for `block.FeesByCoin`
The new test exercises the verbose API path. `GetExplorerBlock`'s population of `block.FeesByCoin` from `skaFeeTotals` (`pgblockchain.go:6577-6594`) and the per-tx `CoinType` populators are still untested. A fixture-based test (block with VAR + SKA1 + SKA2 in `summary.SSFeeTotalsByCoin`) asserting `block.FeesByCoin` content + ordering would lock in the fix and protect against regressions if the source field is renamed/refactored.

#### 4. VAR fee source differs between paths
- API (`apiroutes.go:602-604`): `*summary.MiningFee` — `int64` atoms direct from `BlockDataBasic`.
- Template (`pgblockchain.go:6580-6583`): `int64(block.MiningFee * 1e8)` — `block.MiningFee` is the float64 sum from `getTotalFee(block.Tx) + getTotalFee(block.Treasury) + getTotalFee(block.Revs) + getTotalFee(block.Tickets) + getTotalFee(block.Votes)`.

These compute **different values**, not just different precision: the template path sums VAR fees from all categories of txs in the block; the summary's `MiningFee` may have a narrower scope. VAR's 8-decimal precision fits float64 safely so this isn't a precision bug, but the two paths emitting different numbers under the same "VAR fee" label is worth a follow-up. Cleanest fix: route the template path through `*summary.MiningFee` too.

Not a blocker for this PR.

#### 5. `ChartsUpdated` is now a no-op wrapper
`cmd/dcrdata/internal/explorer/explorer.go:929-934`:

```go
func (exp *explorerUI) ChartsUpdated() {
    exp.pageData.Lock()
    // MixedPercent removed - was used for privacy mix tracking
    exp.pageData.Unlock()
}
```

After the cleanup the function only locks/unlocks, with no work between. It still has callers (`cmd/dcrdata/main.go:971,984`), so it can't simply be deleted, but in this state it's confusing — a reader sees a method named `ChartsUpdated` and expects it does *something* on chart updates. Two options:

- If a future commit will repopulate the body (e.g. recomputing some other home-page derived value), leave a TODO and a tracking issue.
- Otherwise, remove the function and its two call-sites in `main.go`.

The dangling `// MixedPercent removed - was used for privacy mix tracking` line should also go either way; removal context belongs in the commit, not the code.

#### 6. Minor — explicit nilling of `response.Fees`
`apiroutes.go:626-628`:

```go
if len(response.Fees) == 0 {
    response.Fees = nil
}
```

`encoding/json`'s `omitempty` already treats a `len(0)` map as empty, so this nil-out is redundant. (Same reasoning would apply to `response.TotalSent`, which is left as a possibly-empty initialized map.) Either drop the nil-out or apply it consistently to both fields. Pure stylistic.

### Overall

The SKA-fee path now has a real source and template/API consistency. Once #1 (stale doc) and #2 (semantic confirmation) are addressed, this PR is in good shape against #223. #3-#6 are recommended follow-ups that don't have to land in this PR.

edshav (commented):
## Re-review of new commits (`c4dd68cf`, `75ccbd18`)

Both follow-ups land cleanly and address items #1 and #2 from my prior review. No new defects spotted. From my side this is good to merge against #223; the remaining items are non-blocker follow-ups that can land separately.

### Addressed
- **Stale godoc removed** (`c4dd68cf`, `explorer/types/explorertypes.go:280`): the leftover `FeesByCoinFromMiningFee` doc lines are gone; `FeesByCoinFromAmounts` now reads cleanly. ✓
- **Fee semantic documented** (`75ccbd18`):
  - `FeesByCoinFromAmounts` godoc now states VAR fees come from `MiningFee` (sum of VAR tx fees in the block) and SKA fees come from `SSFeeTotalsByCoin` (total SKA atoms distributed via `TxTypeSSFee` per coin type, i.e., fees collected from SKA outputs being spent). ✓
  - `BlockInfo.FeesByCoin` field has the same explanation, so a reader hitting either the producer or the consumer gets the semantic. ✓
  - One-line comment in `getBlockVerbose` (`apiroutes.go:608`) ties the API path to the same source. ✓

The wording is accurate to the code: VAR via `*summary.MiningFee` int64 atoms, SKA via `summary.SSFeeTotalsByCoin`. Ambiguity around "what does `Fees` mean" from my prior review is resolved.

### Carried over (non-blocker, OK to defer)

These were flagged in the prior review and remain open. None are blockers for #223; listing for tracking:

- **#3** — No template-path test for `block.FeesByCoin` (only the verbose-API path is covered by `TestBlockVerbose_SKAFeesFromSSFeeTotals`). A fixture test on `GetExplorerBlock` would lock in the source-of-truth choice.
- **#4** — VAR fee source still differs between paths: API uses `*summary.MiningFee` (`int64` atoms), template uses `int64(block.MiningFee * 1e8)` from the float64 sum across categories. Same label, potentially different numbers. Cleanest fix is routing the template path through `*summary.MiningFee` too.
- **#5** — `explorer.go:929-934` `ChartsUpdated` is a lock/unlock no-op with a `// MixedPercent removed - was used for privacy mix tracking` comment. Either repopulate it (with a TODO + tracking issue) or drop the function and its two callers in `main.go:971,984`.
- **#6** — Explicit `response.Fees = nil` when empty is redundant given `omitempty`; `response.TotalSent` is left as-is. Stylistic only.

### Overall

The new commits do exactly what was asked and nothing extra. From my side this PR satisfies the acceptance criteria of #223. **Recommend merge** once any other reviewers are satisfied; the four carried-over items are good fodder for a follow-up PR.

edshav (approved):

PR #241, opened 2026-05-08
docs: add address page multi-coin implementation plan

yanchenko-igor:
Sequential 18-task plan for transforming address page from VAR-only to VAR + SKA{n} multi-coin model.

Phases:
- Types & Mocks (tasks 1-5)
- Query Layer (tasks 6-9)
- API & Handler (tasks 10-13)
- Caching (tasks 14-15)
- Testing (tasks 16-18)

### DIFF
--- a/docs/implementation/address-multicoin-plan.md
+++ b/docs/implementation/address-multicoin-plan.md
@@ -2,7 +2,9 @@
 
 ## Overview
 
-Transform the address page `/address/{address}` from single-coin (VAR-only) to multi-coin (VAR + SKA{n}) model, following the contracts defined in wiki specs.
+**Backend-only plan** - Transform the address page `/address/{address}` from single-coin (VAR-only) to multi-coin (VAR + SKA{n}) model, following the contracts defined in wiki specs.
+
+Templates (`address.tmpl`, `extras.tmpl`), Stimulus controller (`address_controller.js`), and CSV download UX are **out of scope** and covered separately.
 
 **Source References:**
 - `wiki/specs/address-overview/spec.md` - Cross-feature contract
@@ -13,29 +15,38 @@ Transform the address page `/address/{address}` from single-coin (VAR-only) to m
 - `wiki/code-analysis/address/charts.impact.md` - Charts analysis
 - `wiki/code-analysis/address/transactions.impact.md` - Transactions analysis
 
+## Not in Scope
+
+The following are handled in separate plans:
+- Template changes (`address.tmpl`, `extras.tmpl`, `addresstable.tmpl`)
+- Frontend controller (`address_controller.js`)
+- CSV download UX (frontend)
+
 ## Key Principles
 
 1. **SKA precision**: All SKA values pass as strings end-to-end - no `float64` conversion
-2. **Computation in Go**: All atom arithmetic happens in Go backend (`big.Int` for SKA)
+2. **Computation in Go**: All atom arithmetic happens in Go backend (`*big.Int` for SKA)
 3. **URL contract**: `?coin=N` is the single new query param shared between charts and transactions
 4. **ActiveCoins**: Central concept - determines which coins to display per address
+5. **Precomputation in Go**: Per-coin totals (`TotalReceived`, `TotalOutputs`, `TotalInputs`) computed in Go so template never does SKA string arithmetic
 
 ## Dependencies
 
-- `db/dbtypes/types.go` - AddressInfo, AddressBalance, AddressTx, AddressRow
+- `db/dbtypes/types.go` - AddressInfo, AddressBalance, AddressTx, AddressRow, ChartsData
 - `db/dcrpg/queries.go` - retrieveAddressBalance, new queries
-- `db/dcrpg/pgblockchain.go` - AddressData, AddressHistory, TxHistoryData
+- `db/dcrpg/pgblockchain.go` - AddressData, AddressHistory, TxHistoryData, UnconfirmedTxnsForAddress
 - `db/dcrpg/internal/addrstmts.go` - SQL statements
 - `db/cache/addresscache.go` - Caching logic
-- `cmd/dcrdata/internal/api/apiroutes.go` - Chart endpoints
+- `cmd/dcrdata/internal/api/apiroutes.go` - Chart endpoints, CSV download
 - `cmd/dcrdata/internal/api/apirouter.go` - Route registration
 - `cmd/dcrdata/internal/explorer/explorerroutes.go` - Page handler
 - `cmd/dcrdata/internal/explorer/explorer_test.go` - Test mocks
 - `cmd/dcrdata/internal/api/noop_ds_test.go` - Test mocks
+- `mempool/` package - Unconfirmed transactions per address
 
 ## Implementation Order
 
-### Phase 1: Types & Mocks (Tasks 1-5)
+### Phase 1: Types & Mocks (Tasks 1-6)
 
 #### Task 1: Add ActiveCoins to AddressInfo
 **Files**: `db/dbtypes/types.go`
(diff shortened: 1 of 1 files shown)


edshav (commented):
## Summary

The plan organizes 18 sequential tasks across five phases. As a **backend-only** plan it's a reasonable skeleton, but several spec-mandated backend pieces are missing and a few task descriptions disagree with the spec on load-bearing details. The doc itself should also state its scope explicitly so readers don't expect template/JS work in this plan.

---

## Scope clarity (doc-level)

The Overview says: "Transform the address page `/address/{address}` from single-coin (VAR-only) to multi-coin (VAR + SKA{n}) model." This reads as full-stack, but the plan intentionally covers backend only — templates (`address.tmpl`, `extras.tmpl`), the Stimulus controller (`address_controller.js`), and the CSV download UX are out of scope and will land in a separate plan.

Please surface this in the doc — e.g. amend the Overview to say "**backend-only plan**; templates and frontend covered separately" and add a "Not in scope" section that points to the future doc. Otherwise an implementer will close this milestone and assume the page is done, when in fact `data-address-balance` (`address.tmpl:15`), `Credit VAR` / `Debit VAR` literals (`extras.tmpl:222-236`), the sstxcommitment heuristic at `extras.tmpl:277`, the `customizedFormatter` `DCR` strings (`address_controller.js:84,130,140`), etc. are still live.

---

## Backend gaps (within stated scope)

1. **No mempool / unconfirmed work.** Summary spec §3.5 + §6.4 require `pgb.mp.UnconfirmedTxnsForAddress` (`db/dcrpg/pgblockchain.go:2587-2592`) to return `map[uint8]int64` and a new `NumUnconfirmedByCoin` field on `AddressInfo`. Transactions spec §3.8 requires the mempool overlay (`pgblockchain.go:2632-2634, 2693-2695`) to populate `CoinType` / `SKAValue` per row instead of routing everything through `dcrutil.Amount(...).ToCoin()`. Both are backend changes (in `mempool/` package and `pgblockchain.go`) and absent from the plan.

2. **Server-side cumulative balance is missing.** Charts spec §3.3.1 makes a deliberate, load-bearing decision: the running balance for the `balance` view is computed in Go via `*big.Int.Add` inside `parseRowsSentReceived` (`db/dcrpg/queries.go:4139-4163`) and shipped as `Balance []float64` (VAR) / `BalanceAtoms []string` (SKA), so JS never does BigInt arithmetic. Plan's Task 11 says "return per-coin chart series" — that's not enough; the server-side cumsum needs an explicit task, otherwise an implementer will leave it on the JS side and the spec invariant is violated.

3. **`dbtypes.ChartsData` extension missing.** Charts spec §5.3: `ChartsData` (`db/dbtypes/types.go:2030-2040`) needs `Balance []float64` for VAR and `BalanceAtoms []string` plus `ReceivedAtoms` / `SentAtoms` / `NetAtoms` for SKA. Plan touches endpoint behavior in Tasks 10–11 but not the type itself.

4. **CSV download (`AddressRowCompact` + handler) is out.** This is backend, not frontend. Transactions spec §5.5 requires `AddressRowCompact` (`db/dbtypes/types.go:1299-1309`) to grow `CoinType` / `SKAValue` and the CSV serializer (`apiroutes.go:1729-1775`) to emit `coin_type` + a single string `amount` column instead of `dcrutil.Amount(...).ToCoin()`. The route is also cached 180s, so the cache key needs `?coin=`. None of this is in the plan, and it's all backend.

5. **`AddressTable` XHR handler parity.** Transactions spec §8: `?coin=` must be parsed in **both** `AddressPage` (`explorerroutes.go:1534`) and `AddressTable` (`:1654`), and `mergedTxnCount` (`pgblockchain.go:2300-2331`) needs the same coin filter — otherwise the XHR refresh paginates over a different row set than the initial SSR. Plan only mentions parsing once in Task 12.

6. **Per-coin precomputed totals — captured weakly.** Summary spec §3.3-3.4 mandates `TotalReceived` per `CoinBalance` (= `TotalSpent + TotalUnspent`) and scalar `TotalOutputs` / `TotalInputs` on `AddressBalance` so the (eventual) template doesn't add SKA strings — there's deliberately no `addSkaAtoms` template helper. Task 3 lists `TotalOutputs`/`TotalInputs` but doesn't explain the rationale; Task 2 has `TotalReceived` but no note that the addition happens in Go. Without the rationale a future PR will skip the precompute and force the template author to add a helper that the spec is explicitly avoiding.

7. **Charts vs table default behavior on empty `?coin=` differs (handler concern).** Plan Task 10: "default to VAR (0) if not specified". Spec actually says: chart endpoints default to **first member of `ActiveCoins`** (VAR if present, else first SKA), table list endpoint **does not filter** (returns all coins). These are different handler-side defaults on the same URL key.

8. **Summary handler must ignore `?coin=`.** Address-overview §4 + summary §1: the summary card always shows the full per-coin breakdown regardless of `?coin=`. That's a handler invariant (`AddressData` populates the `Coins` map for all `ActiveCoins` regardless of filter) — worth calling out in Task 12 so an implementer doesn't apply the filter universally.

9. **Stake metrics gating to VAR.** Summary spec §3.6: `FromStake` / `ToStake` are computed off VAR atoms only. `db/dcrpg/queries.go:1518-1524` is currently coin-flattened, producing a meaningless ratio on mixed addresses. This is a backend fix and isn't in the plan.

10. **Task 5 mocks list is missing `DevBalance`.** Per `explorer.go:84-86`, `mockDataSource` implements `AddressHistory`, `AddressData`, **and** `DevBalance`. If `AddressBalance` shape changes, all three must update.

---

## Smaller issues

11. **Task 13's reasoning for removing Fiat is wrong.** Plan says: "fiat removed as SKA has no price model." Summary spec §3.7 is explicit that fiat is removed because **no coin in the network has a real market** — including VAR. Worth fixing because someone reading the plan will assume fiat survives for VAR-only addresses; it does not. The handler change (drop `exp.xcBot.Conversion(...)` at `explorerroutes.go:1617` and the `FiatBalance` field on the inline payload at `:1543`) is also concrete enough to be its own task.

12. **`CoinBalance` field shape is ambiguous.** The struct mixes `TotalSpent int64` and `TotalSpentSKA string`, and which one is real depends on `CoinType`. Worth documenting that explicitly in the plan (or splitting into two structs) — readers shouldn't have to reverse-engineer which field is meaningful.

13. **Open Questions are mostly already answered by the specs.**
    - "Keep deprecated `TotalUnspent`/`TotalSpent`?" — overview §7.1 has the new `Coins map[uint8]*CoinBalance` shape; the flat fields go.
    - "What to do with addresses with empty `ActiveCoins`?" — summary spec §3.8 specifies the empty state explicitly.
    - "Cache backward compatibility?" — `AddressCache` is in-memory (rebuilt per process), so this isn't actually a concern.

---

## Suggestions

- Amend the Overview / add a "Not in scope" section calling out that templates, controller JS, and CSV-UX are covered in a separate plan.
- Add explicit tasks for: mempool overlay coin-awareness, `UnconfirmedTxnsForAddress` per-coin map, **CSV schema change in the handler**, server-side balance cumsum in `parseRowsSentReceived`, `ChartsData` extension, `mergedTxnCount` coin filter, `AddressTable` XHR coin parsing parity, VAR-only stake metric calculation.
- Resolve the three Open Questions in the plan itself by quoting the relevant spec sections — they're already decided.
- Verification steps could be more behavioral (e.g. "fetch `/api/address/{addr}/amountflow/day?coin=1` for a SKA1-funded address and confirm `BalanceAtoms` is non-zero and monotonically updating") rather than "code compiles" — but that's a nice-to-have, not blocking.

edshav (commented):
## Follow-up — re-review of the updated plan

Thanks for the rewrite. All twelve concerns from yesterday's review are addressed; the plan is now well-aligned with the wiki specs and is implementable. A handful of small follow-ups below — all minor, none blocking.

### Concerns from prior review — confirmed resolved

| # | Concern | Where it landed |
|---|---------|----------------|
| Scope | Plan reads as full-stack | Overview prefix "**Backend-only plan**" + "Not in Scope" section ✅ |
| 1 | Mempool / `UnconfirmedTxnsForAddress` per-coin | New Phase 3 — Tasks 13, 14 ✅ |
| 2 | Server-side cumsum in `parseRowsSentReceived` | Task 11, with explicit `*big.Int.Add` rationale ✅ |
| 3 | `dbtypes.ChartsData` extension | Task 5 ✅ |
| 4 | CSV (`AddressRowCompact`, serializer, cache key) | Task 17 ✅ |
| 5 | `AddressTable` XHR parity + `mergedTxnCount` filter | Task 18 ✅ |
| 6 | Precomputed totals rationale | Task 2 + Task 3 — both call out "no `addSkaAtoms` template helper" ✅ |
| 7 | Charts vs table default on empty `?coin=` | Task 15 (charts → first ActiveCoins) + Task 19 (table → no filter) ✅ |
| 8 | Summary ignores `?coin=` | Task 20 ✅ |
| 9 | Stake metrics VAR-only | Task 12 ✅ |
| 10 | `DevBalance` in mocks list | Task 6 ✅ |
| 11 | Fiat reasoning (no market for *any* coin, not just SKA) | Task 21 — wording matches summary spec §3.7 ✅ |
| 12 | Open Questions already answered by specs | "Answered Questions" section closes all three ✅ |

### Small follow-ups

1. **Task 17 doesn't mention parsing `?coin=` in the CSV handler.** It only mentions adding the param to the cache key. Per transactions spec §5.5, the CSV route needs `parseAddressParams` (or an equivalent local parse) before it can filter or key the cache by coin. The handler isn't shared with `explorerroutes.go`'s `AddressPage`, so this won't fall out of Task 19. Worth a one-line addition: "parse `?coin=` from the request and pass to query".

2. **`CoinBalance`: dual-shape design is now documented but still implicit.** Task 2 explains that `TotalSpent`/`TotalUnspent` are meaningful when `CoinType==0` and the SKA fields are meaningful otherwise — good. Just flagging that this is now an *invariant the caller must respect at every read site*, not a self-checking type. If you want it to be self-checking, splitting into `VARCoinBalance` and `SKACoinBalance` (or a tagged-union helper method) would catch wrong-field reads at compile time. Documentation-only is fine if the team prefers it; just want to be explicit that the trade-off is now baked in.

3. **Task 5 — verify there's no name collision on `Balance`.** The plan adds `Balance []float64` to `ChartsData` but says "Existing fields..." in the comment. Charts spec §5.3 also asks for a new `Balance []float64` parallel to `Received`/`Sent`/`Net`. Quickly check `db/dbtypes/types.go:2030-2040` at impl time to make sure no existing `Balance` field is being shadowed.

4. **"Risk Areas" still lists "Template changes required — High".** Now stale — the plan's scope explicitly excludes templates. Either drop the row or rephrase to "Coordinate with frontend team on contract change" so it doesn't contradict the Not-in-Scope section.

5. **Task ordering note (nit).** Task 16 (`TxHistoryData` coin filter) consumes the new `Balance`/`BalanceAtoms` fields populated by Task 11. Worth adding a one-liner to Task 11: "prerequisite for Task 16" — would make the dependency obvious to whoever picks up Phase 4 in isolation.

### Spot-check against wiki specs

Confirmed the plan's spec citations are accurate:
- Summary spec §3.3 (precomputed `TotalReceived`), §3.4 (scalar `TotalOutputs`/`TotalInputs`), §3.6 (VAR-only stake metrics), §3.7 (fiat removal reason) — all match.
- Charts spec §3.3.1, §5.3 (`ChartsData` extension, server-side cumsum) — match.
- Transactions spec §5.5 (CSV schema), §8 (XHR parity) — match.

LGTM with the five small follow-ups above. None require another review round — happy for these to be addressed in implementation PRs.

edshav (approved):

yanchenko-igor:
Will be merged with implementation #258

PR #258, opened 2026-05-11
Implement Multi-Coin Backend Support for Address Page

yanchenko-igor:
Summary: Implement Multi-Coin Backend Support for Address Page
This PR transforms the address page backend from a single-coin (VAR-only) model to a multi-coin model (VAR + SKA{n}), as specified in the implementation plan. It enables the explorer to track, calculate, and serve balance and transaction data for multiple coin types per address while maintaining strict precision for SKA atoms.
Key Changes
1. Core Data Structures (db/dbtypes)
- Multi-Coin Balance: Replaced flat balance fields in AddressBalance with a Coins map[uint8]*CoinBalance.
- CoinBalance Struct: Introduced CoinBalance to handle the dual-nature of assets (int64 atoms for VAR, string atoms for SKA).
- Chart Data Extension: Extended ChartsData to support server-side cumulative balances for SKA (BalanceAtoms, etc.), ensuring the frontend never performs BigInt arithmetic.
- Address Info: Added ActiveCoins to AddressInfo to identify all coin types present at a specific address.
2. Query Layer & Logic (db/dcrpg)
- Coin-Aware Queries: Implemented retrieveAddressCoinTypes and updated retrieveAddressBalance to preserve the coin-type dimension.
- Server-Side Cumsum: Implemented running balance calculations in Go using *big.Int for SKA atoms to prevent precision loss.
- Stake Metrics Fix: Updated stake income/spending ratios to be computed exclusively from VAR coin balances, as stake is VAR-denom uma.
- Coin Filtering: Modified AddressData and TxHistoryData to support the ?coin=N filter, defaulting to the first member of ActiveCoins.
3. Mempool & Helpers
- Coin-Aware Mempool: Updated UnconfirmedTxnsForAddress to return transaction counts per coin type.
- Tx Metadata: Extended TxWithBlockData and PrevOut to carry CoinType and SKAValue throughout the pipeline.
4. API & Explorer Handlers
- Chart API: Added ?coin=N filtering to chart endpoints.
- CSV Exports: Extended CSV logic to include coin_type and output SKA amounts as strings.
- XHR Parity: Ensured AddressTable XHR refreshes respect the same coin filtering as the initial SSR.
- Fiat Removal: Removed all fiat conversion logic (FiatBalance) as no coins in the network currently have a reliable market price.
5. Caching & Performance
- Coin-Aware Cache: Updated AddressCache to store per-coin balance maps and chart series.
- Filtered Caching: Chart data is now cached per-coin to ensure efficient retrieval of filtered views.
6. Backward Compatibility (Hotfix)
- Added temporary mock fields to AddressBalance and AddressPageData to prevent template execution failures (TotalUnspent, FiatBalance, etc.) until the frontend templates are updated.
Verification Results
- Unit Tests: New tests for CoinBalance JSON marshaling and ActiveCoins sorting.
- Integration Tests: Verified AddressData and TxHistoryData with mixed-coin addresses.
- API Tests: Validated ?coin=N filtering and default coin fallback behavior.
- Build: Verified that the dcrdata binary builds without errors.
Related Issues/Specs
- wiki/specs/address-overview/spec.md
- wiki/specs/address-summary/spec.md
- wiki/specs/address-charts/spec.md
- wiki/specs/address-transactions/spec.md

### DIFF
--- a/docs/implementation/address-multicoin-plan.md
+++ b/docs/implementation/address-multicoin-plan.md
@@ -2,7 +2,9 @@
 
 ## Overview
 
-Transform the address page `/address/{address}` from single-coin (VAR-only) to multi-coin (VAR + SKA{n}) model, following the contracts defined in wiki specs.
+**Backend-only plan** - Transform the address page `/address/{address}` from single-coin (VAR-only) to multi-coin (VAR + SKA{n}) model, following the contracts defined in wiki specs.
+
+Templates (`address.tmpl`, `extras.tmpl`), Stimulus controller (`address_controller.js`), and CSV download UX are **out of scope** and covered separately.
 
 **Source References:**
 - `wiki/specs/address-overview/spec.md` - Cross-feature contract
@@ -13,29 +15,38 @@ Transform the address page `/address/{address}` from single-coin (VAR-only) to m
 - `wiki/code-analysis/address/charts.impact.md` - Charts analysis
 - `wiki/code-analysis/address/transactions.impact.md` - Transactions analysis
 
+## Not in Scope
+
+The following are handled in separate plans:
+- Template changes (`address.tmpl`, `extras.tmpl`, `addresstable.tmpl`)
+- Frontend controller (`address_controller.js`)
+- CSV download UX (frontend)
+
 ## Key Principles
 
 1. **SKA precision**: All SKA values pass as strings end-to-end - no `float64` conversion
-2. **Computation in Go**: All atom arithmetic happens in Go backend (`big.Int` for SKA)
+2. **Computation in Go**: All atom arithmetic happens in Go backend (`*big.Int` for SKA)
 3. **URL contract**: `?coin=N` is the single new query param shared between charts and transactions
 4. **ActiveCoins**: Central concept - determines which coins to display per address
+5. **Precomputation in Go**: Per-coin totals (`TotalReceived`, `TotalOutputs`, `TotalInputs`) computed in Go so template never does SKA string arithmetic
 
 ## Dependencies
 
-- `db/dbtypes/types.go` - AddressInfo, AddressBalance, AddressTx, AddressRow
+- `db/dbtypes/types.go` - AddressInfo, AddressBalance, AddressTx, AddressRow, ChartsData
 - `db/dcrpg/queries.go` - retrieveAddressBalance, new queries
-- `db/dcrpg/pgblockchain.go` - AddressData, AddressHistory, TxHistoryData
+- `db/dcrpg/pgblockchain.go` - AddressData, AddressHistory, TxHistoryData, UnconfirmedTxnsForAddress
 - `db/dcrpg/internal/addrstmts.go` - SQL statements
 - `db/cache/addresscache.go` - Caching logic
-- `cmd/dcrdata/internal/api/apiroutes.go` - Chart endpoints
+- `cmd/dcrdata/internal/api/apiroutes.go` - Chart endpoints, CSV download
 - `cmd/dcrdata/internal/api/apirouter.go` - Route registration
 - `cmd/dcrdata/internal/explorer/explorerroutes.go` - Page handler
 - `cmd/dcrdata/internal/explorer/explorer_test.go` - Test mocks
 - `cmd/dcrdata/internal/api/noop_ds_test.go` - Test mocks
+- `mempool/` package - Unconfirmed transactions per address
 
 ## Implementation Order
 
-### Phase 1: Types & Mocks (Tasks 1-5)
+### Phase 1: Types & Mocks (Tasks 1-6)
 
 #### Task 1: Add ActiveCoins to AddressInfo
 **Files**: `db/dbtypes/types.go`
@@ -52,20 +63,23 @@ Add `ActiveCoins []uint8` field to `AddressInfo` struct:
 #### Task 2: Create CoinBalance struct
 **Files**: `db/dbtypes/types.go`
 
-Define new per-coin balance structure:
+Define new per-coin balance structure. Note: For VAR (CoinType=0), use `TotalSpent`/`TotalUnspent` (int64). For SKA (CoinType>0), use `TotalSpentSKA`/`TotalUnspentSKA` (string). The meaningful field depends on CoinType.
+
 ```go
 type CoinBalance struct {
     CoinType         uint8
     NumSpent         int64
     NumUnspent       int64
-    TotalSpent       int64        // VAR atoms (1e8)
-    TotalUnspent     int64        // VAR atoms (1e8)
-    TotalSpentSKA    string       // SKA atoms as string (1e18)
-    TotalUnspentSKA  string       // SKA atoms as string (1e18)
-    TotalReceived    int64        // VAR atoms
-    TotalReceivedSKA string       // SKA atoms as string
+    TotalSpent       int64        // VAR atoms (1e8) - meaningful when CoinType == 0
+    TotalUnspent     int64        // VAR atoms (1e8) - meaningful when CoinType == 0
+    TotalSpentSKA    string       // SKA atoms as string (1e18) - meaningful when CoinType > 0
+    TotalUnspentSKA  string       // SKA atoms as string (1e18) - meaningful when CoinType > 0
+    TotalReceived    int64        // VAR atoms - precomputed: TotalSpent + TotalUnspent
+    TotalReceivedSKA string      // SKA atoms - precomputed: TotalSpentSKA + TotalUnspentSKA
 }
 ```
+
+**Rationale**: TotalReceived is precomputed in Go (not template) so the template never does SKA string arithmetic. See Summary spec §3.3.
 
 **Verification**: Struct compiles, JSON marshaling produces correct SKA string format
 
@@ -76,9 +90,12 @@ type CoinBalance struct {
 
 Replace flat balance with per-coin structure:
 - Add `Coins map[uint8]*CoinBalance` to `AddressBalance`
-- Keep deprecated `TotalUnspent`, `TotalSpent` for compatibility or remove
-- Add `FromStake`, `ToStake` (VAR-only)
-- Add computed `TotalOutputs int64`, `TotalInputs int64`
+- Remove flat `TotalUnspent`, `TotalSpent` (replaced by per-coin structure)
+- Add `FromStake`, `ToStake` (computed from VAR only - see Task 21)
+- Add precomputed `TotalOutputs int64 = Σ(NumSpent + NumUnspent)` across all coins
+- Add precomputed `TotalInputs int64 = Σ(NumSpent)` across all coins
+
+**Rationale**: Precomputed totals exist so template doesn't loop over Coins map. This avoids creating a template helper for SKA string addition - the spec deliberately prevents this.
 
 **Breaking Change**: Requires handler/template updates in same PR
 
@@ -97,22 +114,45 @@ Replace flat balance with per-coin structure:
 
 ---
 
-#### Task 5: Update test mocks
+#### Task 5: Add ChartsData type extension
+**Files**: `db/dbtypes/types.go` (around line 2030-2040)
+
+Extend `ChartsData` for server-side cumulative balance (per Charts spec §5.3):
+```go
+type ChartsData struct {
+    // Existing fields...
+    // Add for per-coin:
+    Balance      []float64        // VAR running balance (float64, 8 decimals)
+    BalanceAtoms []string         // SKA running balance (string, 18 decimals)
+    ReceivedAtoms []string        // SKA received cumulative (string)
+    SentAtoms    []string         // SKA sent cumulative (string)
+    NetAtoms     []string         // SKA net cumulative (string)
+}
+```
+
+**Rationale**: Running balance computed in Go via `*big.Int.Add` inside `parseRowsSentReceived`, not in JS. See Charts spec §3.3.1.
+
+**Verification**: Type compiles, JSON marshaling correct
+
+---
+
+#### Task 6: Update test mocks
 **Files**:
 - `cmd/dcrdata/internal/explorer/explorer_test.go`
 - `cmd/dcrdata/internal/api/noop_ds_test.go`
 
-- Update `AddressHistory` return types
-- Add empty `ActiveCoins` default
+- Update `AddressHistory` return types (per Tasks 1-5)
 - Update `AddressData` mock signatures
+- Update `DevBalance` mock (implements AddressHistory, AddressData, AND DevBalance per explorer.go:84-86)
+- Add empty `ActiveCoins` default
 
 **Verification**: Test files compile
 
 ---
 
-### Phase 2: Query Layer (Tasks 6-9)
+### Phase 2: Query Layer (Tasks 7-12)
 
-#### Task 6: New coin type query
+#### Task 7: New coin type query
 **Files**: `db/dcrpg/internal/addrstmts.go`, `db/dcrpg/queries.go`
 
 Create `MakeSelectAddressCoinTypes` statement:
@@ -129,7 +169,7 @@ Add `retrieveAddressCoinTypes(ctx, db, address) ([]uint8, error)`
 
 ---
 
-#### Task 7: Fix retrieveAddressBalance
+#### Task 8: Fix retrieveAddressBalance
 **Files**: `db/dcrpg/queries.go` (around line 1453-1529)
 
 Modify to preserve coin type dimension:
@@ -142,18 +182,19 @@ Modify to preserve coin type dimension:
 
 ---
 
-#### Task 8: Fix ReduceAddressHistory
+#### Task 9: Fix ReduceAddressHistory
 **Files**: `db/dbtypes/types.go` (around line 2380+)
 
 - When `addrOut.CoinType != 0`, use SKA string fields
 - Build per-coin totals in new structure instead of flat sum
 - Don't use `dcrutil.Amount(addrOut.Value).ToCoin()` for SKA
+- Precompute `TotalReceived` and `TotalReceivedSKA` per coin
 
 **Verification**: Mixed addresses don't sum incorrectly
 
 ---
 
-#### Task 9: Update AddressHistory
+#### Task 10: Update AddressHistory
 **Files**: `db/dcrpg/pgblockchain.go` (around line 2445-2496)
 
 - Call `retrieveAddressCoinTypes`
@@ -164,9 +205,56 @@ Modify to preserve coin type dimension:
 
 ---
 
-### Phase 3: API & Handler (Tasks 10-13)
+#### Task 11: Server-side cumulative balance
+**Files**: `db/dcrpg/queries.go` (around line 4139-4163 - `parseRowsSentReceived`)
 
-#### Task 10: Add ?coin= to chart API endpoints
+Modify to compute running balance in Go via `*big.Int.Add`:
+- For VAR: compute running balance as `[]float64` (existing pattern)
+- For SKA: compute running balance as `[]string` using `*big.Int`
+- Populate `BalanceAtoms`, `ReceivedAtoms`, `SentAtoms`, `NetAtoms` in response
+
+**Rationale**: Per Charts spec §3.3.1 - JS never does BigInt arithmetic. Balance cumsum must be in Go.
+
+**Verification**: Chart endpoint returns monotonically updating balance series
+
+---
+
+#### Task 12: Fix stake metrics VAR-only
+**Files**: `db/dcrpg/queries.go` (around line 1518-1524)
+
+Compute `FromStake` and `ToStake` from VAR atoms only:
+- Current query is coin-flattened, producing meaningless ratio on mixed addresses
+- Filter to `coin_type = 0` when computing stake percentages
+
+**Verification**: Stake % reflects actual VAR stake activity
+
+---
+
+### Phase 3: Mempool (Tasks 13-14)
+
+#### Task 13: Unconfirmed transactions per-coin
+**Files**: `db/dcrpg/pgblockchain.go` (around line 2587-2592), `mempool/` package
+
+- Modify `pgb.mp.UnconfirmedTxnsForAddress` to return `map[uint8]int64` (per Summary spec §3.5, §6.4)
+- Add `NumUnconfirmedByCoin` field to `AddressInfo`
+
+**Verification**: Unconfirmed count broken down by coin type
+
+---
+
+#### Task 14: Mempool overlay coin-awareness
+**Files**: `db/dcrpg/pgblockchain.go` (around line 2632-2634, 2693-2695)
+
+- Modify mempool overlay to populate `CoinType` and `SKAValue` per row
+- Don't route everything through `dcrutil.Amount(...).ToCoin()`
+
+**Verification**: Mempool transactions include coin type
+
+---
+
+### Phase 4: API & Handler (Tasks 15-21)
+
+#### Task 15: Add ?coin= to chart API endpoints
 **Files**:
 - `cmd/dcrdata/internal/api/apirouter.go`
 - `cmd/dcrdata/internal/api/apiroutes.go`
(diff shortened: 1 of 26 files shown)


docs/implementation/address-multicoin-plan.md
```diff
@@ -0,0 +1,464 @@
...
+Replace flat balance with per-coin structure:
+- Add `Coins map[uint8]*CoinBalance` to `AddressBalance`
+- Remove flat `TotalUnspent`, `TotalSpent` (replaced by per-coin structure)
+- Add `FromStake`, `ToStake` (computed from VAR only - see Task 21)
+- Add precomputed `TotalOutputs int64 = Σ(NumSpent + NumUnspent)` across all coins
+- Add precomputed `TotalInputs int64 = Σ(NumSpent)` across all coins
+
+**Rationale**: Precomputed totals exist so template doesn't loop over Coins map. This avoids creating a template helper for SKA string addition - the spec deliberately prevents this.
+
+**Note on CoinBalance dual-shape**: `TotalSpent`/`TotalUnspent` (int64) are meaningful when `CoinType==0`; `TotalSpentSKA`/`TotalUnspentSKA` (string) are meaningful when `CoinType>0`. This is an **invariant every read site must respect** - wrong-field reads are silent bugs. If stricter type safety is desired, split into `VARCoinBalance` and `SKACoinBalance`.
```
edshav, line 100: **Blocking: `CoinBalance` dual-shape**

This is the one plan-level concern I'd like resolved before we move to code review.

The unified `CoinBalance` struct with two parallel field sets (`TotalSpent int64` / `TotalUnspent int64` valid only when `CoinType==0`, and `TotalSpentSKA string` / `TotalUnspentSKA string` valid only when `CoinType>0`) creates a category of silent bugs that the type system would otherwise catch:

1. `CoinBalance{CoinType: 1, TotalUnspent: 5_000_000_000}` is a perfectly valid Go value that means nothing — 5 billion VAR atoms on a SKA1 row. Every read site has to remember the discriminator; mocks in tests can silently produce wrong data.
2. `omitempty` on both branches makes `TotalSpent: 0` (a legitimate zero VAR balance) indistinguishable from "not applicable for this SKA row." JSON consumers can't tell from the payload alone.
3. Task 24 says "test `CoinBalance` JSON marshaling (SKA as string)" — singular. To meaningfully test the dual-shape invariant you'd need separate VAR-shape, SKA-shape, and illegal-mixed-shape tests. That's testing what a type system would give you for free.

This struct will end up everywhere — handlers, cache, JSON API, templates, tests. Refactoring it after merge means touching all of those, plus any external consumers that already deserialize the JSON. So now is the cheapest moment to decide.

Two options I'd accept:

- **(a) Split into two types** — `VARCoinBalance` and `SKACoinBalance`, surfaced through a sealed interface or `Coins map[uint8]CoinBalance` where `CoinBalance` is the interface. Awkward at the map level (type assertions) but eliminates the wrong-field-read class of bug entirely.
- **(b) Keep the unified struct, but add `func (c *CoinBalance) Assert()`** that panics on shape mismatch, and call it from every constructor and every read site that branches on `CoinType`. Turns silent bugs into loud ones during dev and tests, with minimal API churn.

The plan itself flags this as Medium risk and mentions option (a) as a possibility — could you make a call here and update the plan to commit to one approach? If you've already considered this and have a reason for the unified struct, happy to hear it.

docs/implementation/address-multicoin-plan.md
```diff
@@ -0,0 +1,464 @@
...
+
+**Verification**: Page renders without fiat errors
+
+---
+
+### Phase 5: Caching (Tasks 22-23)
+
+#### Task 22: Extend address cache key
+**Files**: `db/cache/addresscache.go`
+
```
edshav, line 357: **Cache key shape — distinguish the two caches**

Worth distinguishing the two caches in `db/cache/addresscache.go` rather than treating them uniformly:

- **Balance cache** (`AddressCacheItem.balance`) — stays keyed by address. The `Coins map[uint8]*CoinBalance` lives inside the cached value, so one entry per address covers all coins.
- **Chart cache** (`AddressCacheItem.history`) — keyed by `(address, coin_type)` because charts spec §3.3.1 produces per-coin series.

The current bullet "Add coin type to cache key" reads like both get the same treatment, which would produce N entries per address for what should be one. Not a correctness issue, just suboptimal — but worth nailing down before implementation. Also: please add a test for cache invalidation on new-block arrival (the Risk table flags it Medium and Task 25 doesn't cover it).

edshav (commented):
**Heads-up on review approach**

Thanks for the detailed plan doc and PR description — including the implementation plan in the same PR makes this much easier to review thoroughly. Since the plan wasn't reviewed before implementation, we'd like to do this in **two passes**:

1. **Plan review first** (this round) — validate the plan against the wiki specs it cites and the project invariants in `CLAUDE.md`.
2. **Code review next round** — once the plan is settled, review the diff against the corrected plan, looking for unplanned scope and line-level issues.

Most of the plan reads well and tracks the wiki specs. I have one blocking concern (the `CoinBalance` shape — inline comment below), one performance suggestion, and a list of items we'll verify against the actual code in Pass 2 rather than debate in the plan.

---

**Items we'll verify in code review (Pass 2)**

Rather than debate these on the plan doc, we'll verify them against the actual code in the next review pass. Flagging them here so you know what to expect:

1. **Default-coin behavior (Tasks 15, 19).** Overview §4 specifies: `?coin=` empty → charts default to VAR if VAR ∈ ActiveCoins, else first ActiveCoins. Plan wording ("first member of ActiveCoins") is correct only because the list is sorted ascending. We'll check the handler matches the spec literally.
2. **SKA SUM in amountflow SQL (Task 11).** Charts spec §3.3.3 requires `received_atoms` / `sent_atoms` / `net_atoms` / `balance_atoms` per bin. The existing `selectAddressAmountFlowByAddress` SUMs `value INT8` only — we'll check that a SKA-aware path using `SUM(ska_value::numeric)` was added (charts spec §5.2).
3. **No stake series for SKA (Task 16).** Charts spec §3.3.2: SKA endpoints must not return `Tickets` / `Votes` / `RevokeTx` series. We'll check the SQL omits `tx_type IN (1,2,3)` for SKA.
4. **`mergedTxnCount` coin filter (Task 18).** Transactions §3.5: pagination count must match the filtered row set. We'll check `retrieveMergedTxnCount` SQL has `WHERE coin_type = $K`.
5. **`FillAddressTransactions` (`pgblockchain.go:2746`).** Transactions §5.2 explicitly calls this out as needing `CoinType` / `SKAValue` for DB-sourced rows (not just mempool). Not in any plan task — we'll check the code handles it.

If any of these are already implemented correctly, the plan just needs a wording tweak. If not, we'll raise them as code-level issues in the next round.

---

A few smaller items that don't block plan approval but are worth noting:

- The `Backward Compatibility (Hotfix)` fields in `AddressBalance` / `AddressPageData` (`TotalUnspent`, `FiatBalance`, etc.) aren't reflected in the plan — Task 3 says "remove flat fields, breaking change requires template updates in same PR," but the PR keeps them as TODO-marked stubs because templates are out of scope. That's a reasonable transitional choice given the split, but the plan should be honest that these shims exist and should be tracked for removal in the follow-up frontend PR. Could you add a short "Transitional shims" section to the doc listing them and which PR will remove them?
- The diff includes `cmd/dcrdata/dcrdata` (the built binary) as an added file. Could you drop that and add it to `.gitignore` if it isn't already?

edshav (commented):
# PR #258 review — multi-coin backend for address page

**Scope of this review.** Backend-only changes (23 files, +1636/−341). The `docs/implementation/address-multicoin-plan.md` doc added by this PR was **not** read or evaluated — this review judges the functional Go/SQL/test code against the canonical specs in `wiki/specs/address-{overview,summary,charts,transactions}/spec.md` only. Front-end / template work is out of scope (no `views/*.tmpl` or `public/js/` changes in this PR).

---

## 🚫 Blockers (correctness, must fix before merge)

### B1. SKA balances are never correctly aggregated from the DB
**Spec:** overview §7.1 / summary §6.1 / C1 (precision)
**Where:** [db/dcrpg/internal/addrstmts.go:218](db/dcrpg/internal/addrstmts.go#L218) + [db/dcrpg/queries.go:1503-1574](db/dcrpg/queries.go#L1503-L1574) + [db/dcrpg/queries.go:4880-4901](db/dcrpg/queries.go#L4880-L4901)

`SelectAddressSpentUnspentCountAndValue` SUMs `addresses.value` (int64 VAR atoms) — it ignores `addresses.ska_value` (TEXT). For SKA rows the `value` column is 0 in this schema (overview §2 calls this out as the existing pre-PR pathology). `retrieveAddressBalance` then feeds that 0 into `incrementStringAtoms(...)` which short-circuits on zero, leaving every SKA `TotalSpentSKA` / `TotalUnspentSKA` as `""`. Even if `value` were non-zero, `incrementStringAtoms` multiplies by `1e10` to "convert VAR atoms (1e8) to SKA atoms (1e18)" — that's an invented number, not the on-chain SKA amount stored in `ska_value`.

Net effect: the summary card's SKA balances are always blank or fabricated. The very thing this PR claims to fix.

Fix: extend the GROUP-BY SQL to additionally return `SUM(ska_value::numeric)::text` (or a parallel statement), and aggregate it as a `*big.Int` in Go. Delete `incrementStringAtoms`.

### B2. `ReduceAddressHistory` overwrites SKA atoms row-by-row instead of summing
**Spec:** overview §7.2, transactions §5.2, C1
**Where:** [db/dbtypes/types.go:2469-2473](db/dbtypes/types.go#L2469-L2473) and [:2503-2522](db/dbtypes/types.go#L2503-L2522)

```go
} else {
    tx.ReceivedTotalSKA = addrOut.SKAValue
    receivedSKAByCoin[coinType] = addrOut.SKAValue   // assignment, not sum
}
```
Two SKA receiving rows ⇒ last value wins, prior values discarded. Spent SKA atoms are never accumulated at all (the spent path only updates `sentByCoin[ct]` when `coinType == 0`). Down at the second pass:

```go
cb.TotalUnspentSKA = sum.String()  // sum was only set from receivedSKAByCoin[ct]; spent never added in
```
…so SKA `TotalUnspentSKA` always equals SKA `TotalReceivedSKA`. The test `TestReduceAddressHistory_MixedCoin` (single SKA row) does not exercise the bug.

Fix: keep `receivedSKAByCoin` and `spentSKAByCoin` as `map[uint8]*big.Int`, `.Add(...)` per row, format strings only when writing into `CoinBalance`.

### B3. `parseRowsSentReceived` has no SKA pathway — charts for SKA are broken
**Spec:** charts §3.3.1, §3.3.3, §5.3
**Where:** [db/dcrpg/queries.go:4189-4222](db/dcrpg/queries.go#L4189-L4222)

The function only emits `Received/Sent/Net/Balance []float64` from `SUM(value)`. For `?coin=N>0` it runs the same float pipeline against (zeroed) `value` rows, returning all-zero series. The spec-mandated `ReceivedAtoms/SentAtoms/NetAtoms/BalanceAtoms []string` are never populated. The `dbtypes.ChartsData` shape was extended correctly, but no producer fills it.

Test `TestAddressChartAPI_CoinFiltering` hides this: it uses a hand-rolled mock `TxHistoryData` that hard-codes the expected SKA JSON instead of going through `retrieveTxHistoryByAmountFlow` → `parseRowsSentReceived`.

Fix: add a `parseRowsSentReceivedSKA` (or a coin-aware switch in the same function) that SUMs `ska_value::numeric` as text and accumulates via `*big.Int.Add` (precedent: SKA `coin-supply/{N}`). Update `selectAddressAmountFlowByAddress` to also project SKA atoms text when `$2 != 0`.

### B4. The address SQL filter is mandatory, not optional — default `?coin=` returns VAR-only
**Spec:** overview §4 (semantics), transactions §3.5 ("если ?coin= отсутствует — таблица показывает все монеты")
**Where:** [db/dcrpg/internal/addrstmts.go:190-256](db/dcrpg/internal/addrstmts.go#L190-L256) + [cmd/dcrdata/internal/middleware/apimiddleware.go:817-839](cmd/dcrdata/internal/middleware/apimiddleware.go#L817-L839)

Every address SQL now requires `coin_type = $2` (mandatory). `CoinCtx` middleware silently defaults missing/invalid `?coin=` to `0`. Result: when a user opens an SKA-only address with no `?coin=`, the table is empty (filtered to VAR-only) and the API returns no rows. The spec says absent `?coin=` ⇒ "All coins"; only the chart card defaults to VAR.

Fix: either (a) make the filter optional in SQL — generate two query variants, no `coin_type=$K` predicate when filter absent; or (b) add a different default policy for transactions vs charts at the handler layer. Option (a) is cleaner.

### B5. The non-VAR coin filter silently falls back to "first ActiveCoin" instead of "no filter"
**Spec:** transactions §3.5 ("значение не принадлежит ActiveCoins — фильтр отсутствует")
**Where:** [cmd/dcrdata/internal/explorer/explorerroutes.go:1889-1908](cmd/dcrdata/internal/explorer/explorerroutes.go#L1889-L1908)

```go
if !found {
    coinType = addrData.ActiveCoins[0]
    addrData, err = exp.dataSource.AddressData(...)   // extra round trip
}
```
Spec says treat as filter-absent. Behaviour here: surreptitiously rewrite to a different coin without telling the user, with a second DB round-trip on every miss. Also impossible to implement "All coins" because of B4.

Fix: when `coinType ∉ ActiveCoins`, treat as absent (no SQL filter — see B4). No re-fetch.

### B6. Summary card cannot show all coins because `Balance.Coins` is filtered to the selected coin
**Spec:** overview §4 ("Карточка summary `?coin=` игнорирует — она всегда показывает полный пер-монетный разрез"), summary §3.2
**Where:** [db/dcrpg/pgblockchain.go:2575-2588](db/dcrpg/pgblockchain.go#L2575-L2588)

```go
filteredBalance := &dbtypes.AddressBalance{Address: address, Coins: make(...)}
if coinBal, ok := balance.Coins[coinType]; ok {
    filteredBalance.Coins[coinType] = coinBal
}
addrData.Balance = filteredBalance
```
`addrData.Balance` is what feeds the summary card. After this filter, only the selected coin survives — the multi-coin summary list cannot be rendered. The whole point of overview §4 is that summary ignores `?coin=`.

Fix: keep the full `balance.Coins` map on `addrData.Balance`. Filtering for the transactions table belongs at the SQL/row layer (B4), not on the balance struct.

### B7. Unconfirmed counter double-counts outpoints instead of distinct tx hashes
**Spec:** summary §3.5 / overview §5 ("numUnconfirmed updates live")
**Where:** [mempool/monitor.go:571-599](mempool/monitor.go#L571-L599)

Original returned `int64(len(outs.TxnsStore))` — number of distinct mempool txs touching the address. New code increments `numByCoin[coinType]++` per outpoint in two separate loops (PrevOuts + Outpoints). A single tx with 3 funding outs to the same address contributes 3; a tx that both spends and funds the address contributes from both loops. The aggregate `addrData.NumUnconfirmed = sum(numByCoin)` is therefore inflated.

Fix: track seen txhash-per-coin (e.g. `map[uint8]map[chainhash.Hash]struct{}`) and convert to count at the end. Match original "distinct tx" semantics, just per-coin.

### B8. Mempool SKA balance overlay overwrites instead of accumulating
**Spec:** transactions §5.6, summary §6.4, C1
**Where:** [db/dcrpg/pgblockchain.go:2722-2734](db/dcrpg/pgblockchain.go#L2722-L2734) and [:2812-2820](db/dcrpg/pgblockchain.go#L2812-L2820)

```go
addrData.Balance.Coins[coinType].TotalUnspentSKA = receivedSKA   // assignment
...
addrData.Balance.Coins[coinType].TotalSpentSKA = f.SKAValue       // assignment
```
Same flaw as B2 but in the overlay code. Two mempool SKA receives clobber to the latest. Must be a `*big.Int.Add` against the existing string. Also note `numReceived++` is gated to `coinType == 0` but `addrData.Balance.TotalOutputs++` is unconditional — the two counters drift for SKA, and the final VAR-only block at [:2828-2837](db/dcrpg/pgblockchain.go#L2828-L2837) adds `numReceived` to `Coins[0]` again, re-using counters in inconsistent ways.

Fix: route the overlay through the same accumulation primitive as B2; track per-coin received/sent/numReceived/numSent in one structure; write to `Coins[ct]` exactly once at the end.

### B9. `cmd/dcrdata/dcrdata` is a 38 MB committed Linux binary
**Hygiene**
**Where:** [cmd/dcrdata/dcrdata](cmd/dcrdata/dcrdata)

Not gitignored: `.gitignore:7` excludes `/dcrdata` (repo root) and `cmd/dcrdata/monetarium-explorer`, but not this exact path. Remove the file and add `cmd/dcrdata/dcrdata` to `.gitignore` (or extend the existing rules to cover it).

---

## ⚠️ Should-fix (spec deviations, design smells)

### S1. `AddressBalance` carries flat "mock" fields alongside the new `Coins` map
**Spec:** overview §7.1 (replace flat with per-coin)
**Where:** [db/dbtypes/types.go:2376-2384](db/dbtypes/types.go#L2376-L2384)

```go
// TODO: Remove these fields once frontend is updated for multi-coin support
TotalUnspent int64 `json:"total_unspent"`
TotalSpent   int64 `json:"total_spent"`
NumSpent     int64 `json:"num_spent"`
NumUnspent   int64 `json:"num_unspent"`
```
And [types.go:2511-2517](db/dbtypes/types.go#L2511-L2517) where they're populated from `coins[0]` only. The spec is explicit: flat fields go away. Keeping them as VAR-shadowed shims pollutes the JSON shape and tempts downstream code (Insight API already keys off `Coins[0]` — fine — but external consumers of `AddressBalance` JSON will now see ambiguous "is `total_unspent` the only balance or just VAR?" data).

Fix: delete the shim fields. If a specific consumer (Insight? home page?) still needs flat VAR, give it a dedicated getter, not duplicated JSON-tagged fields. The PR title says "backend support"; the wire format is part of that contract.

### S2. `FiatBalance` half-removed instead of removed
**Spec:** summary §3.7, §6.3 ("Поле FiatBalance *exchanges.Conversion удаляется из inline-структуры AddressPageData")
**Where:** [cmd/dcrdata/internal/explorer/explorerroutes.go:1544](cmd/dcrdata/internal/explorer/explorerroutes.go#L1544) and [:1632](cmd/dcrdata/internal/explorer/explorerroutes.go#L1632)

The field stayed, retyped as `interface{}` set to `nil`, with a TODO. Spec says delete it. Keeping `interface{}` nil is worse than deleting — it changes the template-visible type and risks `{{if .FiatBalance}}` constructions evaluating differently. The `xcBot.Conversion(...)` call is removed (✅), so just remove the struct field too.

### S3. SKA chart Tx-Types query still emits all 5 series for SKA
**Spec:** charts §3.3.2, §5.2 ("Для SKA: только SentRtx + ReceivedRtx — стейк-серии не отображаются")
**Where:** [db/dcrpg/internal/addrstmts.go:319-325](db/dcrpg/internal/addrstmts.go#L319-L325)

The SQL still counts `tx_type = 1/2/3` (Tickets/Votes/SSRtx) even when filtered to a SKA coin. In practice those are 0 for SKA (chain invariant), so it's wasted bandwidth not wrong. Either suppress the stake CASE clauses when `$2 != 0` (two SQL variants) or document in the spec that the over-emission is fine and the FE will drop the series. Picking one or the other matters for the FE contract.

### S4. CSV `amount` format is variable-precision for VAR
**Spec:** transactions §3.7 (CSV unified `amount` string, 8dp for VAR, 18dp for SKA)
**Where:** [cmd/dcrdata/internal/api/apiroutes.go:1802-1808](cmd/dcrdata/internal/api/apiroutes.go#L1802-L1808)

```go
amount = strconv.FormatFloat(dcrutil.Amount(r.Value).ToCoin(), 'f', -1, 64)
```
`-1` precision yields the shortest representation that round-trips — `50` becomes `50`, not `50.00000000`. SKA path returns raw atoms (`"1000000000000000000"`) which is fine but is the integer-atoms form, not the 18-dp decimal form the spec calls for. Pick one CSV contract and apply uniformly; the spec leans toward decimal-formatted strings for both, but if the team prefers atoms-strings (simpler), update the spec, not the implementation in passing.

Fix: VAR → `%.8f`. SKA → either `formatAtomsAsCoinString(s, ct)` for decimal, or document that `amount` is atoms when `coin_type != 0`.

### S5. Duplicated `TotalOutputs/TotalInputs` computation in two layers
**Where:** [db/dcrpg/queries.go:1567-1573](db/dcrpg/queries.go#L1567-L1573) (in `retrieveAddressBalance`) and [db/dbtypes/types.go:2522-2526](db/dbtypes/types.go#L2522-L2526) (in `ReduceAddressHistory`) and again [db/dcrpg/pgblockchain.go:2480](db/dcrpg/pgblockchain.go#L2480) (in the fallback branch)

Three independent computations of the same scalar from different inputs. Easy to drift. Pick one. The `retrieveAddressBalance` one is the canonical source (whole-address aggregate); the others should just copy it.

### S6. `nonMergedTxnCount` returns 0 when the requested coin is absent — UI sees empty table
**Spec:** transactions §3.5/§8 (SSR/XHR parity, sensible defaults)
**Where:** [db/dcrpg/pgblockchain.go:2349-2367](db/dcrpg/pgblockchain.go#L2349-L2367)

```go
coinBal := bal.Coins[coinType]
if coinBal == nil { return 0, nil }
```
Coupled with B4/B5 this produces a degenerate UX: ask for a coin the address doesn't have ⇒ counter says 0 ⇒ pagination breaks. Should be handled at the handler layer (treat absent coin as "no filter") rather than here, but the current early-return at least needs `B5`'s fallback to fire before it's hit.

### S7. SKA stake-percentage gate is correct, but `addrInfo.AmountSent/Unspent` fallback uses VAR-only data unconditionally
**Where:** [db/dcrpg/pgblockchain.go:2472-2492](db/dcrpg/pgblockchain.go#L2472-L2492)

The fallback branch that runs when `addrInfo, _, _ = ReduceAddressHistory(...)` succeeds builds `Coins[0]` from `addrInfo.AmountSent/AmountUnspent` — VAR-only. If the address has SKA-only history, `ActiveCoins=[N≠0]` but `Balance.Coins[0]` is zero-populated regardless and other coins are missing. The `else` path in `AddressHistory` (with `addressRows` non-empty) loses SKA balances from `addrInfo.Balance.Coins` because this branch replaces `balance` with a freshly-built VAR-only struct.

Fix: when this fallback fires, copy `addrInfo.Balance.Coins` wholesale (it already has the per-coin map from `ReduceAddressHistory`), and only overwrite/add `Coins[0]` if the VAR aggregate from `addrInfo.Amount*` differs (which it shouldn't).

### S8. SQL filter still mandatory for `mergedTxnCount` path even when `IsMerged` is false
**Where:** [db/dcrpg/pgblockchain.go:2329-2340](db/dcrpg/pgblockchain.go#L2329-L2340)

`mergedTxnCount` always passes `coinType` to merged-count SQL. With B4 that means SKA-only mempool counts go to zero for `?coin=0`. Same root cause as B4 — fix it once at the SQL layer and this collapses.

---

## 🧹 Nits

### N1. Log line still says "DCR"
**Where:** [db/dcrpg/pgblockchain.go:2502-2504](db/dcrpg/pgblockchain.go#L2502-L2504)

```go
log.Infof("%s: %d spent totalling %f DCR, %d unspent totalling %f DCR", ...)
```
The codebase still has `dcrd`-flavored naming elsewhere (CLAUDE.md flags this as intentional load-bearing), but `DCR` here is a stale coin label not a package name. Replace with `VAR` or drop the suffix.

### N2. `cacheMetrics.rowHit()` is incremented before the rows-nil check
**Where:** [db/cache/addresscache.go:887-895](db/cache/addresscache.go#L887-L895)

```go
ac.cacheMetrics.rowHit()
rows, _ := aci.Rows()
if rows == nil { return nil, nil }
```
A miss counted as a hit. Cosmetic but it skews metrics. Move `rowHit()` past the nil check; on nil, call `rowMiss()`.

### N3. `Rows(addr, coinType)` filters cached rows in-memory on every call
**Where:** [db/cache/addresscache.go:893-899](db/cache/addresscache.go#L893-L899)

For addresses with many rows this defeats the cache hot-path. Acceptable for v1 but worth a note for follow-up.

### N4. Empty test stub `TestAddressDataCoinFiltering`
**Where:** [cmd/dcrdata/internal/api/address_api_test.go:30-34](cmd/dcrdata/internal/api/address_api_test.go#L30-L34)

```go
func TestAddressDataCoinFiltering(t *testing.T) {
    // Test logic will be added here later.
}
```
Either fill in or delete. Empty test passes and creates the illusion of coverage. Same for the dangling `// ...` and trailing test scaffolding in `TestAddressCSV_CoinFiltering`.

### N5. `TestReduceAddressHistory_MixedCoin` has a single SKA row
**Where:** [db/dbtypes/types_test.go:276-345](db/dbtypes/types_test.go#L276-L345)

Add a second SKA-funding row and a SKA-spending row. Without them, B2 (overwrite vs sum) is invisible — and would have prevented this PR from reaching review.

### N6. `rpcutils/rpcclient.go` UnconfirmedTxnsForAddress returns nil map
**Where:** [rpcutils/rpcclient.go:553-606](rpcutils/rpcclient.go#L553-L606)

Documented as "callers should use MempoolMonitor for per-coin counts" — but the `MempoolAddressChecker` interface now requires returning the map. Callers that hold the RPC variant (e.g. Insight `MempoolAddressChecker` consumers) get a nil map, and any `map[coin]++` style downstream code will panic. Sweep for callers; either populate from `Tx.MsgTx().TxOut[i].CoinType` here or remove the RPC implementation altogether if it's no longer reachable.

### N7. `pgblockchain.go:AddressData` references pre-existing `addrErr` / `err` confusion
**Where:** [db/dcrpg/pgblockchain.go:2521-2525](db/dcrpg/pgblockchain.go#L2521-L2525)

```go
_, addrType, addrErr := txhelpers.AddressValidation(...)
if addrErr != nil && !errors.Is(err, txhelpers.AddressErrorNoError) {
    return nil, err   // 'err' is the outer-scope nil
}
```
Pre-existing in `develop` — not caused by this PR but it surfaces in the diff. Worth fixing while the code is open.

---

## Snapshot

| Severity | Count |
|---|---|
| Blocker | 9 (correctness on SKA balances, charts, filter semantics, mempool counting, plus binary hygiene) |
| Should-fix | 8 |
| Nit | 7 |

The PR's plumbing (signatures, middleware, cache key, mocks, `AddressInfo.ActiveCoins`, `CoinBalance` shape, CSV schema, fiat removal) is largely in place and correct. The **arithmetic over SKA atoms** — which is the entire point of the multi-coin redesign — is not. Until B1–B3 + B8 are fixed, the backend cannot serve a correct SKA summary card or SKA chart, regardless of how anything renders downstream. B4–B6 turn the `?coin=` semantics inside out vs the spec. B7 inflates unconfirmed counts.

Recommended next step: fix the SKA aggregation primitives (one shared `bigAtomAdd` helper used by `retrieveAddressBalance`, `ReduceAddressHistory`, and the mempool overlay), then extend test coverage in `types_test.go` to a 2-SKA-receive + 1-SKA-spend fixture so B2 cannot regress silently.

edshav (approved):

PR #266, opened 2026-05-15
db/dcrpg: fix multicoin balance not showing on address page

yanchenko-igor:
The shortcut path in AddressHistory was fetching only coin type 0 (VAR) instead of all coin types when calling AddressTransactionsAll. This caused SKA coin balances to be missing from the cached balance.

Fix: Pass dbtypes.CoinTypeAll (255) instead of 0 to retrieveAddressTxns in AddressTransactionsAll so all coin types are fetched for the balance computation.

### DIFF
--- a/cmd/dcrdata/internal/api/apiroutes.go
+++ b/cmd/dcrdata/internal/api/apiroutes.go
@@ -55,7 +55,7 @@ type DataSource interface {
 	GetBlockByHash(context.Context, string) (*wire.MsgBlock, error)
 	SpendingTransaction(ctx context.Context, fundingTx string, vout uint32) (string, uint32, error)
 	SpendingTransactions(ctx context.Context, fundingTxID string) ([]string, []uint32, []uint32, error)
-	AddressHistory(ctx context.Context, address string, N, offset int64, txnType dbtypes.AddrTxnViewType) ([]*dbtypes.AddressRow, *dbtypes.AddressBalance, error)
+	AddressHistory(ctx context.Context, address string, N, offset int64, txnType dbtypes.AddrTxnViewType, coinType uint8) ([]*dbtypes.AddressRow, *dbtypes.AddressBalance, error)
 	FillAddressTransactions(ctx context.Context, addrInfo *dbtypes.AddressInfo) error
 	AddressTransactionDetails(ctx context.Context, addr string, count, skip int64,
 		txnType dbtypes.AddrTxnViewType) (*apitypes.Address, error)
--- a/cmd/dcrdata/internal/api/noop_ds_test.go
+++ b/cmd/dcrdata/internal/api/noop_ds_test.go
@@ -26,7 +26,7 @@ func (noopDS) SpendingTransaction(_ context.Context, _ string, _ uint32) (string
 func (noopDS) SpendingTransactions(_ context.Context, _ string) ([]string, []uint32, []uint32, error) {
 	return nil, nil, nil, nil
 }
-func (noopDS) AddressHistory(_ context.Context, _ string, _, _ int64, _ dbtypes.AddrTxnViewType) ([]*dbtypes.AddressRow, *dbtypes.AddressBalance, error) {
+func (noopDS) AddressHistory(_ context.Context, _ string, _, _ int64, _ dbtypes.AddrTxnViewType, _ uint8) ([]*dbtypes.AddressRow, *dbtypes.AddressBalance, error) {
 	return nil, nil, nil
 }
 func (noopDS) FillAddressTransactions(_ context.Context, _ *dbtypes.AddressInfo) error { return nil }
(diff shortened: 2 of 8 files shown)


edshav (approved):
## Summary

LGTM with one follow-up observation. The one-line fix is correct, well-scoped, and addresses a real bug on the multi-coin balance shortcut path.

## Correctness of the fix

The change in `db/dcrpg/pgblockchain.go:1755` replaces a positional `0` (which the parameter takes as "filter to coin type 0", i.e. VAR-only) with the sentinel `dbtypes.CoinTypeAll` (255, "no filter"). The dispatch lives in `retrieveAddressTxns` at `db/dcrpg/queries.go:1874-1880`:

```go
func retrieveAddressTxns(..., coinType uint8) (..., error) {
    if coinType == dbtypes.CoinTypeAll {
        return retrieveAddressTxnsStmtAll(...)        // SelectAddressLimitNByAddressAll: no coin_type predicate
    }
    return retrieveAddressTxnsStmt(..., coinType)     // SelectAddressLimitNByAddress: WHERE coin_type=$2
}
```

So pre-fix, `AddressTransactionsAll` was misnamed in practice — it returned VAR-only rows. The fix aligns behavior with the doc comment ("retrieves all non-merged main chain addresses table rows") and with the function name.

## Impact trace (why this matters)

`AddressTransactionsAll` is the sole producer for the address-row cache populated in `updateAddressRows` (`pgblockchain.go:2192-2221`), which is consumed by three callers:

1. **`AddressRowsCompact`** (`pgblockchain.go:2266-2308`) — filters in-memory by `r.CoinType == coinType`. With VAR-only rows in the cache, requests for any SKA coin returned an empty slice.
2. **`AddressRowsMerged`** (`pgblockchain.go:2225-2262`) — same source cache; merging over VAR-only rows silently dropped SKA activity.
3. **`AddressHistory` shortcut** (`pgblockchain.go:2474-2495`) — when `len(addressRows) < N && offset == 0 && txnView == AddrTxnAll`, it computes balance via `ReduceAddressHistory(addressRows)` directly from the cached slice. With VAR-only rows, `balance.Coins` ended up missing SKA entries entirely. This is the visible symptom the PR describes.

The non-shortcut balance path (`AddressBalance` → `retrieveAddressBalance` via SQL aggregation, `pgblockchain.go:2502`) was already multi-coin correct, which is why the bug only surfaced for addresses where `rows < limit`.

The wiki note on `CoinTypeAll` semantics also confirms 255 is the intended sentinel for table/counts pipelines, distinct from the chart pipeline that collapses 255→0.

## Observations (non-blocking)

1. **Sibling function carries the same pattern.** `AddressTransactionsAllMerged` (`pgblockchain.go:1770`) calls `retrieveAddressMergedTxns`, which hard-codes `coinType=0` at `queries.go:1886`. It feeds `AddressRowsMerged`, used by the Insight API (`insightapi.go:81`). Likely the same bug shape for the merged Insight surface — worth a follow-up (out of scope here, since merged queries don't currently have a `*All` variant in `addrstmts.go`).
2. **No regression test.** `pgblockchain_test.go:70` exercises an empty address; `TestMergeRows` at line 98 exercises a non-empty address but doesn't assert per-coin presence. A targeted integration test against a SKA-bearing address would have caught this; adding one is bounded by the available DB fixtures.
3. **Optional doc nit.** The comment at `pgblockchain.go:1746-1749` ("hard limit of 3 million rows...4x the count for the treasury address...") is VAR-only sizing. Not blocking; just a callout that the bound's rationale predates multi-coin.

## Verification done locally

- Diff is a single line; no other call sites of `AddressTransactionsAll` exist outside `db/dcrpg/`.
- `dbtypes.CoinTypeAll` is already imported and used in `pgblockchain.go` at lines 2354, 2374, 2619, 3327-3328 — no new imports needed.
- Branched off `develop` per repo conventions.

edshav (approved):
## Re-review of commit 8b5c3a1b (against prior approved review)

The core one-line fix remains correct and shippable — approving. The follow-up commit only partially addresses the earlier observations; noting these as **non-blocking follow-ups**, not merge blockers.

### Obs #1 — sibling merged path: comment is half-right
The new note in `AddressTransactionsAllMerged` correctly clarifies that `SelectAddressMergedViewAll` has no `coin_type` predicate (`WHERE address=$1` only), so the merged path is **not** the same VAR-only-filter bug — good.

But "the offset (0) here is correct" mis-targets the original concern (the hard-coded `coinType=0` *inside* `retrieveAddressMergedTxns`, queries.go:1886) and obscures a separate **pre-existing** latent bug: `retrieveAddressTxnsStmt` always binds `(address, coinType, N, offset)` → `$1..$4`, but `SelectAddressMergedView` uses `$2`=LIMIT / `$3`=OFFSET. Net effect on the merged path: `LIMIT $2(=coinType=0) OFFSET $3(=N=3000000)` ≈ `LIMIT 0`. Not introduced by this PR and out of scope here, but the comment asserting "correct" will mislead future readers. Suggest dropping/correcting sentence 2 and opening a separate follow-up issue for the merged Insight path (AddressRowsMerged → insightapi.go:81).

### Obs #2 — regression test asserts nothing
The block added to `TestMergeRows` can't catch the regression it names: `len(coinTypes)==0` is unreachable (rows already `Fatalf`'d if nil; `CoinType` is always set), and the multi-coin case is only `t.Logf`. If the fix regressed to `coinType=0`, this test still passes. To actually guard it, fail on `len(coinTypes) <= 1` against a fixture address known to hold SKA rows. Follow-up, bounded by available DB fixtures (as the prior review noted).

### Obs #3 — doc nit: not addressed, fine (was optional).

Net: approving the fix; recommend tracking the test hardening and the merged-path param-mismatch as separate issues.

PR #278, opened 2026-05-18
Fix SKA PoW/PoS reward attribution and VAR fee calculation (closes #273)

yanchenko-igor:
Summary
- Fix SKA Reward Attribution: Replaced the legacy 50/50 split heuristic for SKA rewards with authoritative on-chain markers ("MF" for miners, "SF" for stakers) found in TxType=7 transactions.
- Correct VAR Fee Calculation: Removed the 0-cap in ComputeVoteVARReward to properly account for staker payouts, which can result in negative net redistributions.
- Modernize Reward Types: Introduced the SSFeeSplit struct to explicitly track PoW and PoS components of rewards, replacing the legacy string-based totals.
- Update Consumers: Refactored explorer.go and pubsubhub.go to consume the new SSFeeSplit type and removed the divideByTwoSKAFees helper.
Technical Details
- txhelpers/ssfee.go: Updated BlockSSFeeTotals to compute netReward = Sum(Outputs) - Sum(Inputs) and attribute it based on the presence of markers in any transaction output.
- api/rewardtypes/types.go: Defined the SSFeeSplit struct.
- cmd/dcrdata/internal/explorer/explorer.go: Updated Store() and backward search logic to use the new reward types and removed the divide-by-two heuristic.
- pubsub/pubsubhub.go: Updated Store() to handle the new SSFeeSplit structure.


edshav (changes requested):
## Verdict: Request changes

The PR moves in the right direction (introduces `SSFeeSplit`, reads `stake.HasSSFeeMarker`, removes `divideByTwoSKAFees` and the `TODO(monetarium-node)` workarounds, centralizes the split in `txhelpers`). But it does **not** implement the core algorithm the linked issue and `docs/pow-pos-reward-classification-problem.md` explicitly specify, leaves a VAR requirement unmet, deletes regression tests the spec said to update, and pollutes the root module's dependency graph.

---

### Blocking 1 — SKA marker attribution uses tx-level net instead of summing the actual MF/SF outputs (contradicts the resolved spec decision)

`txhelpers/ssfee.go` `blockSSFeeTotalsInternal` computes `netReward = Σ(outputs) − Σ(inputs)` for the tx, then sets `isPoS`/`isPoW` booleans from whichever markers appear and adds the **whole** `netReward` to each flagged bucket.

The spec's resolved "No rounding replication needed" decision is explicit: the node already wrote the exact per-output amounts; **summing the actual `"MF"`/`"SF"` outputs reads those exact consensus amounts**. Consequences of the net-based approach:

- **Double-counting**: if one SSFee tx carries both an `"SF"` and an `"MF"` output (50/50 split applied per coin type; node routes per-output at `staketx.go:1390`), both `isPoS` and `isPoW` are true and the entire `netReward` is added to **both** buckets. No test covers a mixed-marker tx, so CI stays green.
- **Wrong quantity even for single-marker txs**: `netReward` subtracts inputs from the sum of all outputs, not the value of the marked output(s); these differ whenever the SSFee tx consolidates inputs.

Please implement per-output classification (classify each `tx.TxOut` by its own `HasSSFeeMarker` and add that output's `SKAValue` to the matching bucket) and add a test with one tx containing both an `"SF"` and an `"MF"` output.

### Blocking 2 — VAR still does not isolate `"SF"` SSFee staker payouts

The issue and the spec's resolved "VAR also needs the marker" decision require isolating `"SF"`-marked SSFee payouts for VAR, not only SKA. This PR keeps the lumped VAR net (`Σ(Outputs − Inputs)` over SSGen/SSRtx/SSFee, stored into `PoS`) and only deletes the negative cap in `ComputeVoteVARReward` — treating the symptom, not the requirement.

This is masked by the rewritten `TestComputeVoteVARReward`, which now hard-codes `latestVarFee := 3.19483536` (positive) with a comment (`feeVAR = totalVARInputs - totalVARReward`) that contradicts the implementation's `Outputs − Inputs` (which yields −319483536 for block 36354). The test no longer derives its input from `BlockSSFeeTotals`.

### Blocking 3 — Regression tests deleted rather than updated

The issue's Tests checklist says to update/re-verify `TestBlockSSFeeTotals_WithType7`, `TestFullPipelineWithType7`, `TestFullPipelineWithSSRtx`, `TestComputeVoteVARReward_NegativeFee`, `TestFullPipelineSSGenReward` against the split return type. The PR deletes all of them, dropping VAR-side net/SSRtx/negative-fee and full-pipeline coverage, replaced by a single single-marker SKA table test. Please restore these as split-aware tests.

### Blocking 4 — Root `go.mod` / `go.sum` pollution (~1,500 spurious lines)

`go.mod` gains direct deps `monetarium-explorer/exchanges` and `monetarium-explorer/gov` at remote pseudo-versions plus a large indirect tree (btcsuite, cockroachdb/pebble, gnark-crypto, prometheus, …); `go.sum` grows ~1,500 lines. No root-module package imports `exchanges`/`gov` and the dependency direction per `CLAUDE.md` is the reverse. `go mod tidy` on this branch removes both direct deps and nets **−1,666 lines** across go.mod/go.sum — a tidy artifact from an environment that resolved sibling modules to remote pseudo-versions. Please regenerate root `go.mod`/`go.sum` cleanly. (Dependent-module tidy churn in `db/dcrpg`, `gov`, `pubsub/democlient` looks fine.)

---

### Non-blocking

- Explorer ↔ pubsub drift partially reintroduced: `explorer.go` reads `latestVarFee` via `big.Float`, `pubsubhub.go` via `float64(split.PoS.Int64())` — same quantity, two formulations. Unify via a shared helper.
- `SKAPoWRewards` fallback swaps the `/2` heuristic for a 100%-PoW heuristic rather than deriving from `"MF"` markers; call this out explicitly if intentional.
- Broken godoc: a blank line was inserted into the `explorerDataSource` doc comment in `explorer.go`, splitting it.
- Unrelated churn: `BlockTransactionCounts` moved within `apitypes.go`; `defaultAddressRows` comment reworded (now duplicates `MaxAddressRows`). Consider dropping to keep the diff focused.

---

### Verification performed

- Root-module packages build on the PR branch.
- `go test ./txhelpers/...` passes — but green CI largely reflects deleted coverage and a single-marker-only new test.
- `go mod tidy` confirms the root go.mod/go.sum pollution (−1,666 net lines; both spurious direct deps removed).

edshav (approved):

PR #284, opened 2026-05-20
explorer: implement data contract for /visualblocks rewrite

yanchenko-igor:
Summary
Fixed the data contract for the /visualblocks frontend rewrite to ensure consistency between HTTP and WebSocket transports.
Changes
- Type Updates:
    - Widened TrimmedBlockInfo to include Size, FormattedBytes, CoinFills, ActiveSKACount, RegularCoinCounts, and MaxBlockSize.
    - Updated BlockInfo to export RegularCoinCounts, CoinFills, ActiveSKACount, and MaxBlockSize for WebSocket JSON.
    - Added Voted boolean flag to TrimmedTxInfo.
    - Added MaxBlockSize and TotalSize to TrimmedMempoolInfo.
- Logic Implementation:
    - Implemented ComputeCoinFills in explorertypes.go to calculate block/mempool fill bars using the 10%/90% quota model.
    - Updated MempoolInfo.Trim() to accept maxBlockSize and populate the new contract fields.
    - Updated VisualBlocks HTTP handler to populate TrimmedBlockInfo with contract data.
    - Updated RootWebsocket handler to populate BlockInfo with contract data using shallow copies to prevent shared state mutation.
    - Updated PubSubHub to pass maxBlockSize during mempool trimming.
- Testing:
    - Added visualblocks_contract_test.go to assert that both transports carry the identical data contract and that the Voted flag is correctly set.
    - Updated existing home viewmodel tests to accommodate the Trim() signature change.

### DIFF
--- a/cmd/dcrdata/internal/explorer/dev_indicators.go
+++ b/cmd/dcrdata/internal/explorer/dev_indicators.go
@@ -55,7 +55,7 @@ func buildDevIndicatorScenarios() []devIndicatorScenario {
 	const maxBlock = 100.0
 
 	mkMempool := func(stats map[uint8]types.MempoolCoinStats, issued []uint8) *types.MempoolInfo {
-		fills, totalFill, activeSKA := computeCoinFills(stats, maxBlock, issued)
+		fills, totalFill, activeSKA := types.ComputeCoinFills(stats, maxBlock, issued)
 		mp := &types.MempoolInfo{}
 		mp.MempoolShort.CoinFills = fills
 		mp.MempoolShort.TotalFillRatio = totalFill
--- a/cmd/dcrdata/internal/explorer/explorer.go
+++ b/cmd/dcrdata/internal/explorer/explorer.go
@@ -495,7 +495,7 @@ func (exp *explorerUI) StoreMPData(_ *mempool.StakeData, _ []types.MempoolTx, in
 	if blockchainInfo != nil && blockchainInfo.MaxBlockSize > 0 {
 		maxBlockSize = float64(blockchainInfo.MaxBlockSize)
 	}
-	fills, totalFillRatio, activeSKACount := computeCoinFills(inv.CoinStats, maxBlockSize, issuedSKA)
+	fills, totalFillRatio, activeSKACount := types.ComputeCoinFills(inv.CoinStats, maxBlockSize, issuedSKA)
 	inv.CoinFills = fills
 	inv.MempoolShort.CoinFills = fills
 	inv.MempoolShort.TotalFillRatio = totalFillRatio
@@ -606,7 +606,7 @@ func (exp *explorerUI) Store(blockData *blockdata.BlockData, msgBlock *wire.MsgB
 			maxBlockSize = float64(p.BlockchainInfo.MaxBlockSize)
 		}
 		exp.invsMtx.Lock()
-		fills, totalFillRatio, activeSKACount := computeCoinFills(exp.invs.CoinStats, maxBlockSize, issuedSKA)
+		fills, totalFillRatio, activeSKACount := types.ComputeCoinFills(exp.invs.CoinStats, maxBlockSize, issuedSKA)
 		exp.invs.CoinFills = fills
 		exp.invs.MempoolShort.CoinFills = fills
 		exp.invs.MempoolShort.TotalFillRatio = totalFillRatio
--- a/cmd/dcrdata/internal/explorer/explorerroutes.go
+++ b/cmd/dcrdata/internal/explorer/explorerroutes.go
@@ -345,33 +345,7 @@ func (exp *explorerUI) VisualBlocks(w http.ResponseWriter, r *http.Request) {
 	exp.pageData.RUnlock()
 
 	for _, block := range blocks {
-		// Convert CoinRowData to MempoolCoinStats for ComputeCoinFills
-		stats := make(map[uint8]types.MempoolCoinStats)
-		for _, row := range block.CoinRows {
-			stats[row.CoinType] = types.MempoolCoinStats{
-				Size: int32(row.Size),
-			}
-		}
-		fills, _, activeSKACount := types.ComputeCoinFills(stats, maxBlockSize, issuedSKA)
-
-		trimmedBlock := &types.TrimmedBlockInfo{
-			Time:              block.BlockTime,
-			Height:            block.Height,
-			Total:             block.TotalSent,
-			Fees:              block.MiningFee,
-			Subsidy:           block.Subsidy,
-			Votes:             block.Votes,
-			Tickets:           block.Tickets,
-			Revocations:       block.Revs,
-			Transactions:      types.FilterRegularTx(block.Tx),
-			Size:              block.BlockBasic.Size,
-			FormattedBytes:    block.BlockBasic.FormattedBytes,
-			CoinFills:         fills,
-			ActiveSKACount:    activeSKACount,
-			RegularCoinCounts: types.RegularCoinCountsFromCoinRows(block.CoinRows, block.BlockBasic.Voters, block.BlockBasic.FreshStake, block.BlockBasic.Revocations),
-			MaxBlockSize:      maxBlockSize,
-		}
-
+		trimmedBlock := block.Trim(maxBlockSize, issuedSKA)
 		trimmedBlocks = append(trimmedBlocks, trimmedBlock)
 	}
 
--- a/cmd/dcrdata/internal/explorer/explorerroutes.go
+++ b/cmd/dcrdata/internal/explorer/explorerroutes.go
@@ -335,17 +335,41 @@ func (exp *explorerUI) VisualBlocks(w http.ResponseWriter, r *http.Request) {
 
 	// trim unwanted data in each block
 	trimmedBlocks := make([]*types.TrimmedBlockInfo, 0, len(blocks))
+
+	exp.pageData.RLock()
+	maxBlockSize := float64(exp.pageData.BlockchainInfo.MaxBlockSize)
+	issuedSKA := make([]uint8, 0, len(exp.pageData.HomeInfo.SKACoinSupply))
+	for _, entry := range exp.pageData.HomeInfo.SKACoinSupply {
+		issuedSKA = append(issuedSKA, entry.CoinType)
+	}
+	exp.pageData.RUnlock()
+
 	for _, block := range blocks {
+		// Convert CoinRowData to MempoolCoinStats for ComputeCoinFills
+		stats := make(map[uint8]types.MempoolCoinStats)
+		for _, row := range block.CoinRows {
+			stats[row.CoinType] = types.MempoolCoinStats{
+				Size: int32(row.Size),
+			}
+		}
+		fills, _, activeSKACount := types.ComputeCoinFills(stats, maxBlockSize, issuedSKA)
+
 		trimmedBlock := &types.TrimmedBlockInfo{
-			Time:         block.BlockTime,
-			Height:       block.Height,
-			Total:        block.TotalSent,
-			Fees:         block.MiningFee,
-			Subsidy:      block.Subsidy,
-			Votes:        block.Votes,
-			Tickets:      block.Tickets,
-			Revocations:  block.Revs,
-			Transactions: types.FilterRegularTx(block.Tx),
+			Time:              block.BlockTime,
+			Height:            block.Height,
+			Total:             block.TotalSent,
+			Fees:              block.MiningFee,
+			Subsidy:           block.Subsidy,
+			Votes:             block.Votes,
+			Tickets:           block.Tickets,
+			Revocations:       block.Revs,
+			Transactions:      types.FilterRegularTx(block.Tx),
+			Size:              block.BlockBasic.Size,
+			FormattedBytes:    block.BlockBasic.FormattedBytes,
+			CoinFills:         fills,
+			ActiveSKACount:    activeSKACount,
+			RegularCoinCounts: types.RegularCoinCountsFromCoinRows(block.CoinRows, block.BlockBasic.Voters, block.BlockBasic.FreshStake, block.BlockBasic.Revocations),
+			MaxBlockSize:      maxBlockSize,
 		}
 
 		trimmedBlocks = append(trimmedBlocks, trimmedBlock)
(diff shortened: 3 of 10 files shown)


edshav (commented):
## Summary

The PR widens the type surface in roughly the right places, but it doesn't deliver the contract for the two pieces it most explicitly promises. Two of the eight acceptance criteria in #269 are unmet by the actual data on the wire, the central regression test is a tautology that can't catch the things it claims to assert, and the recommended reuse of `computeCoinFills` was replaced by a duplicate copy.

## Blockers

### 1. Block-side `Voted` is never set → every vote on a mined block renders as "did-not-vote"

`explorer/types/explorertypes.go:1108` sets `Voted: tx.VoteInfo != nil` in `TrimMempoolTx`, but the block path builds vote entries through `trimmedTxInfoFromMsgTx` (`db/dcrpg/pgblockchain.go:6547`) which still only sets `VoteValid`. So for `block.Votes` on both transports, `Voted` stays the Go zero value `false`. Per spec §4 / the issue's mapping `Voted&Valid → YES`, `Voted&!Valid → NO`, `!Voted → did-not-vote`, every real vote on the block tile will be classified as gray.

Fix: set `Voted: txBasic.VoteInfo != nil` in `trimmedTxInfoFromMsgTx`, right next to where `VoteValid` is derived from the same source.

### 2. Coinbase is included in the VAR fill bar — explicit spec violation

The block-side bar in `explorerroutes.go:348-355` and `websockethandlers.go:288-294` feeds `block.CoinRows[ct].Size` straight into `ComputeCoinFills`. Those sizes come from `blockCoinTxStats` (`blockdata/blockdata.go:438-462`), which excludes only stake types (SSGen/SStx/SSRtx) — coinbase is added to VAR. Spec §6.3 and the issue's acceptance criterion both require "coinbase excluded as today", matching today's `FilterRegularTx` behavior on the per-tx segments the bars are replacing.

Fix options: subtract the coinbase size from VAR in the conversion step, or compute the per-coin regular sizes by walking `block.Tx` and skipping `tx.Coinbase` (same filter the JS controller already mirrors).

### 3. `ComputeCoinFills` is a duplicate of `computeCoinFills` — the issue specifically told us not to do this

Issue #269 lists under "Reuse existing types; do not invent new shapes":

> `computeCoinFills()` — `cmd/dcrdata/internal/explorer/explorer.go` (VAR = 10% maxBlockSize, SKA pool = 90% / activeSkaCount).

The PR added a new, exported, near-identical clone in `explorer/types/explorertypes.go:1467-1564`. The original `computeCoinFills` (`cmd/dcrdata/internal/explorer/explorer.go:1100-1209`) is still in use by `Store`, `StoreMPData`, `dev_indicators`, and `templates_test.go`. The home-page indicator and the visualblocks block-tile indicator now run through two copies of the same algorithm, which is the exact failure mode the "same quota model and same struct" wording in the issue exists to prevent.

Fix: either move the function to `explorer/types` and switch the existing callers to it (preferred — the issue's own bullet implies sharing), or keep the unexported one and have the WS/HTTP handlers call into the explorer package. Not both.

### 4. `visualblocks_contract_test.go` doesn't test the contract

`cmd/dcrdata/internal/explorer/visualblocks_contract_test.go` is essentially `f(x) == f(x)`:

- `BlockContractConsistency` builds the "HTTP" `TrimmedBlockInfo` and the "WS" `BlockInfo` by calling `types.ComputeCoinFills` twice with the same inputs and then asserts the results are equal. It never invokes `VisualBlocks` or `RootWebsocket`, never JSON-encodes either payload, and most of the assertions compare struct fields the test itself literal-initialized to the same constant (`Size: 17000` on both sides).
- `MempoolContractConsistency` is a unit test for `Trim`, not a cross-transport check.
- `VoteFlagConsistency` only exercises `TrimMempoolTx`, so it can't detect blocker #1 above.

To meet the issue's "Regression test asserting block **and** mempool payloads carry the full contract (incl. `Voted`) on both transports", the test needs to drive the actual handlers (or at least the actual populating helpers used by them), serialize what each transport sends, and compare the deserialized contract fields field-by-field — including `Voted` on a fixture that exercises the block path, not just the mempool path.

## Quality / smaller issues

### 5. Inline `CoinRowData → MempoolCoinStats` conversion duplicated

The exact same 5-line loop appears in `explorerroutes.go:348-353` and `websockethandlers.go:288-293`. Worth a small helper (e.g. `types.StatsFromCoinRows`) — same drift hazard, smaller scale.

### 6. Two adjacent RLock/RUnlock cycles for two fields of `pageData`

`websockethandlers.go:154-162` and `pubsubhub.go:322-330` both take `pageData.RLock` for `MaxBlockSize`, release, call `Trim`, then take `RLock` again for `NBlockSubsidy`. A single `RLock` block reading both into locals before `Trim` is cleaner and removes the (admittedly narrow) `Store` window between the two acquisitions.

### 7. New `BlockInfo` JSON fields are not `omitempty`

`BlockInfo.ActiveSKACount` and `BlockInfo.MaxBlockSize` (`explorer/types/explorertypes.go:769-770`) have no `omitempty`. The PR only populates them in the `sigNewBlock` path on `/visualblocks`. Any other consumer that JSON-encodes `BlockInfo` (current or future) will now emit `\"active_ska_count\":0,\"max_block_size\":0`. Per `wiki/code-analysis/visualblocks/impact.md` §2, `sigNewBlock` is consumed by `home_latest_blocks_controller.js`, `status_controller.js`, `blocks_controller.js`, `time_controller.js`, and `address_controller.js` — additive but worth a quick eyeball, and either `omitempty` or populating in `Store`/`GetExplorerBlock` would make this honest.

### 8. PR doesn't link issue #269

`closingIssuesReferences` is empty; the body just says "Part of #268". This PR is the deliverable for #269, so add `Refs #269` (or `Closes #269` if you intend to close it on merge) so the link is queryable from `gh`.

---

## What's good

- `MaxBlockSize`, `Size`, `FormattedBytes`, `RegularCoinCounts` are wired through both transports.
- The shallow-copy of `*BlockInfo` in the WS path correctly avoids mutating the memoized `pgb.lastExplorerBlock` pointer (per `patterns.md` §5) — assignments are to the local struct's new fields, not to anything reachable through the shared `*BlockBasic`.
- `Trim` signature change is propagated to all callers including `pubsub/pubsubhub.go`, and the existing `home_viewmodel_test.go` was updated in lockstep.
- The `TrimmedMempoolInfo` addition (`MaxBlockSize` + `TotalSize`) reuses fields already on `MempoolShort` rather than recomputing.

edshav (commented):
Re-reviewed against commits a80b8a23 and 7ac6d11f.

## Resolved

- **1. Block-side `Voted`** — set in `trimmedTxInfoFromMsgTx` (`db/dcrpg/pgblockchain.go:6561`). ✅
- **3. Duplicate `ComputeCoinFills`** — single source now in `explorer/types`; `Store`, `StoreMPData`, `dev_indicators`, `templates_test` all updated. ✅
- **5. Inline `CoinRowData → MempoolCoinStats` bridge** — replaced by `types.StatsFromCoinRows`, used by `BlockInfo.Trim` on both transports. ✅
- **6. Adjacent RLock/RUnlock cycles** — merged in `websockethandlers.go` and `pubsubhub.go`. ✅
- **8. PR ↔ issue link** — `closingIssuesReferences` now contains #269. ✅
- **4. (Mostly)** — `visualblocks_contract_test.go` now drives `BlockInfo.Trim` / `MempoolInfo.Trim`, JSON-marshals the result, and inspects wire fields including `voted`. Real improvement. Remaining gap under D below.

## New issues introduced by the fixes

### A. Blocker: coinbase exclusion uses `bi.Tx[0]`, which isn't the coinbase

`explorer/types/explorertypes.go:724-731`:

```go
coinbaseSize := int32(0)
if len(bi.Tx) > 0 {
    coinbaseSize = bi.Tx[0].TxBasic.Size
}
```

`block.Tx` is sorted descending by `Total` in `pgblockchain.go:6790` (`sortTx(block.Tx)`), and coinbase `Total` is just `subsidy + fees`. Any block containing a regular transfer whose output total exceeds the coinbase reward will have `Tx[0]` be that transfer, not the coinbase. Result:

- VAR fill bar **subtracts a regular tx's bytes** instead of the coinbase's — VAR bar reads too low.
- Coinbase bytes **stay in** the VAR count — spec §6.3 still violated, just less obviously.

Fix: scan for the entry with `TxBasic.Coinbase == true` (mirrors the JS controller's `block.Tx.filter((tx) => !tx.Coinbase)`):

```go
coinbaseSize := int32(0)
for _, tx := range bi.Tx {
    if tx.TxBasic != nil && tx.TxBasic.Coinbase {
        coinbaseSize = tx.TxBasic.Size
        break
    }
}
```

The contract test doesn't catch this because its fixture puts the coinbase first and doesn't set `Total`, so the production sort never reorders the slice. A second fixture with `Tx[0].Total > coinbase.Total` would have caught it.

### B. Blocker: `BlockInfo.RegularCoinCounts` reverted to `json:\"-\"`

Previous commit had `json:\"regular_coin_counts\"`. The new commit at `explorer/types/explorertypes.go:791` sets it back to `json:\"-\"`:

```go
RegularCoinCounts []CoinCount `json:\"-\"`
```

This regresses issue #269's acceptance criterion:

> Block tile carries per-coin **regular-tx counts** (`RegularCoinCounts []CoinCount`, VAR + each active SKA{n}) for tooltips — **JSON-exposed** and added to `TrimmedBlockInfo`

The WS handler at `websockethandlers.go:288-291` only copies `CoinFills`, `ActiveSKACount`, `MaxBlockSize` from `trimmed` onto `blockCopy` — `RegularCoinCounts` is neither copied nor JSON-exposed, so the WS frontend can never read it. The PR description still claims \"Updated BlockInfo to export RegularCoinCounts, CoinFills, ActiveSKACount, and MaxBlockSize for WebSocket JSON,\" but the code now contradicts that.

Fix: restore `json:\"regular_coin_counts,omitempty\"` on `BlockInfo.RegularCoinCounts` (it's already populated by `GetExplorerBlock`, so no extra wiring needed; `omitempty` handles legacy paths that don't populate it).

### C. Blocker: `BlockInfo.Trim` stopped filtering coinbase from `Transactions`

Previous HTTP path: `Transactions: types.FilterRegularTx(block.Tx)`. New `Trim()` (`explorer/types/explorertypes.go:742`): `Transactions: bi.Tx`. The current `visualblocks.tmpl` still renders the per-tx segments row from `.Transactions` (`cmd/dcrdata/views/visualblocks.tmpl:114-123`). Until the frontend rewrite deletes that block, the HTTP-rendered page will now show an extra `<span class=\"block-tx\">` for the coinbase (linking to `/tx/<coinbase-hash>`).

Fix: `Transactions: FilterRegularTx(bi.Tx)` in `Trim()`.

### D. Quality: contract test only checks one transport

`BlockContractWireFormat` marshals `bi.Trim(...)` and inspects that JSON. That's the **HTTP wire** (`TrimmedBlockInfo`). The WS wire is `WebsocketBlock{Block: &blockCopy}` where `blockCopy` is a `BlockInfo` with three fields patched — a different shape with different JSON tags. The test name says \"wire format\" but it tests only one side.

To meet the issue's \"both transports\" requirement, also build a `BlockInfo` that mirrors the WS handler's patch sequence (`blockCopy.CoinFills = trimmed.CoinFills` etc.), marshal that, and assert the WS-side wire shape carries the same contract fields. If you add this, asserting `wire[\"regular_coin_counts\"]` exists would also have caught issue B above.

### E. Minor: nil-deref risk in coinbase scan

`TxBasic` is a pointer embed on `TrimmedTxInfo`. If anything ever produces a `TrimmedTxInfo` with `TxBasic == nil` (defensive against future test fixtures), `bi.Tx[0].TxBasic.Size` panics. Add a nil guard alongside the fix in A.

---

## What's still good

- `StatsFromCoinRows` is a clean place to put the coinbase-exclusion logic — the bug is in the *call site*, not the helper.
- Removing the duplicate `computeCoinFills` and consolidating in `types` is a real reduction in drift surface.
- `BlockInfo.Trim` centralises the trim logic so HTTP and WS now share a single populator — exactly the pattern the dual-transport invariant needs.

edshav (approved):
Reviewed against commit 3f619110.

## Resolved from previous review

- **A. Coinbase identification** — `explorer/types/explorertypes.go:726-731` now scans `bi.Tx` for `TxBasic.Coinbase == true` instead of trusting `Tx[0]`. Matches the JS controller's `block.Tx.filter((tx) => !tx.Coinbase)` semantics. ✅
- **B. `BlockInfo.RegularCoinCounts` JSON tag** — `explorer/types/explorertypes.go:794` restored to `json:"regular_coin_counts,omitempty"`. WS path's `blockCopy := *block` shallow-copies the field, and it's already populated upstream in `db/dcrpg/pgblockchain.go:6640`, so no extra wiring needed. ✅
- **C. `Transactions` filtering** — `explorer/types/explorertypes.go:745` now uses `FilterRegularTx(bi.Tx)`; the HTTP `visualblocks.tmpl` won't render a spurious coinbase span. ✅
- **D. Cross-transport test coverage** — new `BlockWSWireFormat` subtest in `cmd/dcrdata/internal/explorer/visualblocks_contract_test.go` mirrors the WS patch sequence (`blockCopy.CoinFills = trimmed.CoinFills` etc.), marshals, and asserts the four contract fields are present. Asserting `wire["regular_coin_counts"]` here would have caught the B regression. Mostly ✅, with one remaining gap (D′ below).
- **E. Nil-deref guard on `TxBasic`** — `if tx.TxBasic != nil && tx.TxBasic.Coinbase` in the new loop. ✅

The `BlockContractWireFormat` fixture also got the right hardening: regular tx with `Total: 1000000` precedes the coinbase (`Total: 100`) in `bi.Tx`, so a regression to `Tx[0]` would now be caught instead of accidentally passing.

## Remaining (non-blocking) gaps

### D′. WS subtest is a presence smoke test, not a cross-transport equivalence check

`BlockWSWireFormat` only checks that four fields *exist* on the WS wire — not that their values equal the HTTP wire produced from the same `bi`. The issue's "carry the **identical** data contract" wording calls for a value-level comparison: marshal both, compare `coin_fills`, `active_ska_count`, `max_block_size`, `regular_coin_counts` field-by-field. As-is, a future change that, say, computes `CoinFills` differently in the two paths would still pass this test.

Not a blocker — the presence checks cover the B regression class, which was the most acute failure mode. But the test name promises more than it asserts; either rename it (`BlockWSWireFormatPresence`) or extend it to compare values against the HTTP-side wire.

### Nit: WS subtest `bi.Tx` ordering

The WS subtest's `bi.Tx` contains only the coinbase, so the `Tx[0]`-vs-`Coinbase` bug from A wouldn't be exercised on the WS path even if it regressed. If you broaden D′ above, also include a non-coinbase entry so the coinbase scan is meaningfully covered through the WS wire too.

## Verdict

All four blockers from the previous round are addressed. Remaining items are quality, not correctness. Worth a follow-up for D′ since the test is the only line of defense against the kinds of regressions B and C represented.

PR #292, opened 2026-05-21
Fixes the SKA coin supply chart to accurately reflect historical supply, including burn events and issuance peaks.

yanchenko-igor:
Changes
1. Supply Calculation Accuracy
- Excluding Burn Outputs: Updated SelectSKACoinSupplyPerBlock to exclude outputs with script_type = 'nulldata', ensuring burn events correctly decrease the total supply.
- Coinbase Input Handling: Modified the query to exclude coinbase inputs from the subtraction logic, correcting the final total supply to exactly $4.5$ trillion SKA.
2. Chart Granularity & Visibility
- Continuous Data Series: Removed the filter that skipped blocks with zero supply change, providing a smooth, continuous line on the block-level chart.
- Peak Visibility in Daily Bins: Updated the daily aggregation logic in db/cache/charts.go to store the maximum supply reached during a day instead of the final value. This preserves the issuance peak even if a burn occurred within the same 24-hour window.
Verification
- Verified via API that the absolute supply value is now correctly $4,500,000,000,000.00$ SKA.
- Confirmed that bin=block now returns data for every block.
- Confirmed that bin=day now captures and displays the issuance peak.

### DIFF
--- a/cmd/dcrdata/internal/api/apiroutes.go
+++ b/cmd/dcrdata/internal/api/apiroutes.go
@@ -1948,6 +1948,8 @@ func (c *appContext) getTreasuryIO(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, data, m.GetIndentCtx(r))
 }
 
+const skaSupplyStaleHeightThreshold = 10 // approx. 10-20 mins staleness budget
+
 func (c *appContext) ChartTypeData(w http.ResponseWriter, r *http.Request) {
 	chartType := m.GetChartTypeCtx(r)
 	bin := r.URL.Query().Get("bin")
@@ -1960,14 +1962,13 @@ func (c *appContext) ChartTypeData(w http.ResponseWriter, r *http.Request) {
 	if c.charts != nil && cache.IsSKASupplyChart(chartType) {
 		coinType := cache.SkaCoinType(chartType)
 		if coinType > 0 {
-			currHeight, err := c.DataSource.GetHeight(r.Context())
-			reload := false
-			if err != nil {
-				log.Warnf("ChartTypeData: failed to get current height: %v", err)
-			} else {
-				if cachedHeight, ok := c.charts.SKASupplyHeight(coinType); !ok || currHeight-cachedHeight > 10 {
+			reload := !c.charts.SKASupplyExists(coinType)
+			if currHeight, err := c.DataSource.GetHeight(r.Context()); err == nil {
+				if cachedHeight, ok := c.charts.SKASupplyHeight(coinType); !ok || currHeight-cachedHeight > skaSupplyStaleHeightThreshold {
 					reload = true
 				}
+			} else {
+				log.Warnf("ChartTypeData: failed to get current height: %v", err)
 			}
 
 			if reload {
--- a/cmd/dcrdata/internal/api/apiroutes.go
+++ b/cmd/dcrdata/internal/api/apiroutes.go
@@ -1959,9 +1959,21 @@ func (c *appContext) ChartTypeData(w http.ResponseWriter, r *http.Request) {
 	// Check if this is an SKA supply chart and if data needs to be loaded
 	if c.charts != nil && cache.IsSKASupplyChart(chartType) {
 		coinType := cache.SkaCoinType(chartType)
-		if coinType > 0 && !c.charts.SKASupplyExists(coinType) {
-			if err := c.DataSource.LoadSKASupplyForCoin(r.Context(), c.charts, coinType); err != nil {
-				log.Warnf("ChartTypeData: failed to load SKA supply for coin type %d: %v", coinType, err)
+		if coinType > 0 {
+			currHeight, err := c.DataSource.GetHeight(r.Context())
+			reload := false
+			if err != nil {
+				log.Warnf("ChartTypeData: failed to get current height: %v", err)
+			} else {
+				if cachedHeight, ok := c.charts.SKASupplyHeight(coinType); !ok || currHeight-cachedHeight > 10 {
+					reload = true
+				}
+			}
+
+			if reload {
+				if err := c.DataSource.LoadSKASupplyForCoin(r.Context(), c.charts, coinType); err != nil {
+					log.Warnf("ChartTypeData: failed to load SKA supply for coin type %d: %v", coinType, err)
+				}
 			}
 		}
 	}
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -1824,7 +1824,7 @@ func ChartUintsFromStrings(data []string) ChartUints {
 }
 
 // aggregateSKASupply aggregates block-level SKA supply data to daily bins.
-// Returns timestamps (Unix time seconds), block heights, and values for each day.
+// It takes the last recorded supply value of each day to reflect the end-of-day supply.
 func aggregateSKASupply(timestamps []int64, heights []int64, values []string) ([]int64, []int64, []string) {
 	if len(timestamps) == 0 {
 		return nil, nil, nil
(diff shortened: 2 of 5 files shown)


edshav (approved):
## Summary

The core idea is a real correctness fix. The old query returned, for each block, the sum of outputs created in that block **that are still unspent today**. The caller then ran a cumulative sum on that — so the chart's value at block N reflected only outputs that survived to "now," not the actual supply at block N. Replacing it with a per-block delta (`vouts - vins`, with burns and coinbase-style inputs excluded) and accumulating in Go produces an accurate historical supply curve. Math checks out: running_total at tip → `O_unspent`, matching `SelectSKACoinSupply` minus RPC-reported burns.

A few items worth following up on, none of them blocking.

---

## Issues / Discussion

**1. `aggregateSKASupply` now picks max-of-day instead of last-of-day — semantic shift worth a second look**
`db/cache/charts.go:1830`

A daily "supply" chart conventionally shows end-of-day supply, not the intraday peak. Showing the peak on day N when supply ended day N at a lower value due to an intraday burn means:
- Day N shows the peak.
- Day N+1 shows the post-burn value.
- Visually the burn appears to have happened on day N+1, not on day N when it actually occurred.

The block-level bin already shows the true shape. Two alternatives worth considering: keep last-of-day and rely on `bin=block` for peak visibility, or surface peak/issuance as a separate series. If you stick with max-of-day, please make the docstring explicit — currently it just says "aggregates block-level SKA supply data to daily bins" which understates the behavior change.

**2. Test no longer validates the documented behavior**
`db/cache/charts_test.go:434`

`TestAggregateSKASupply_DailyBins` still has `// Verify last-value aggregation` as a comment, and the fixture uses monotonically increasing values (`100, 200, ..., 900`) so `max == last` accidentally. The test passes against the new code but doesn't exercise the new semantics. Please update both the comment and add a case with an intraday burn (e.g. `…, 500, 600, 550, 700, 650`) and assert the peak is preserved. As-is, a regression that reverted the change would still be green.

**3. Outer subquery aggregates the entire `transactions` table to derive `(block_height, block_time)`**
`db/dcrpg/internal/vinoutstmts.go:257-263`

```sql
SELECT block_height, MIN(block_time) as block_time 
FROM transactions 
WHERE is_mainchain AND is_valid 
AND block_height > $1 
GROUP BY block_height
```

This walks the full `transactions` table just to extract a `(height, time)` pair per block. The `blocks` table has `height` and `time` directly — joining against that is `O(blocks)` instead of `O(transactions)`, several orders of magnitude smaller on a mature chain. Since this query runs both during initial chart preload and on-demand per coin type from the API, it's worth not paying tx-table cost. Suggested:

```sql
FROM (
    SELECT height as block_height, time as block_time
    FROM blocks
    WHERE is_mainchain AND height > $1
) t
```

**4. `script_type != 'nulldata'` silently excludes NULL `script_type` rows**
`db/dcrpg/internal/vinoutstmts.go:269`

In SQL, `NULL != 'nulldata'` evaluates to NULL → falsy in `WHERE`. The `script_type` column is `TEXT` with no `NOT NULL` constraint. If any SKA vout ever has a NULL `script_type` (partial sync, schema migration edges, unusual script classes), it would be silently dropped from the supply. Safer form:

```sql
AND vouts.script_type IS DISTINCT FROM 'nulldata'
```

Worth verifying whether SKA vouts are guaranteed to always have a populated `script_type` before deciding.

**5. Implicit invariant: SKA issuance inputs use `prev_tx_hash = \x00…00`**
`db/dcrpg/internal/vinoutstmts.go:278`

The query treats inputs with zero prev_tx_hash as "issuance" and excludes them from the subtraction — matching the convention used by `SelectCoinSupply` for VAR. For VAR there's also a `tx_type = ANY(ARRAY[0,2,6])` filter. The new SKA query has no `tx_type` filter — implicitly assumes that any zero-prev-hash input on the SKA chain represents new issuance regardless of tx_type. If SKA has its own stake-vote / treasury-spend equivalents, this could under- or over-count. Worth a one-line code comment documenting the invariant being relied on, and worth confirming that `monetarium-node` tags SKA issuance inputs the same way.

**6. Column alias `total` is now misleading — query returns a delta**
`db/dcrpg/internal/vinoutstmts.go:256`

Docstring already says "delta" but the alias is still `AS total`. Renaming to `AS delta` would make the SQL self-documenting at every call site. Cosmetic, but the rest of the codebase reads SQL closely.

**7. Verification claim of $4.5T**

The PR description states absolute supply is now exactly $4,500,000,000,000.00 SKA. Cross-checked the math symbolically and it should equal `SelectSKACoinSupply` minus RPC-reported burns at tip. Mostly a callout that this should be re-verified against the new query after merge on mainnet — running totals over deltas can quietly drift from absolute snapshots if `is_mainchain`/`is_valid` flips happen during reorgs and the chart isn't rebuilt.

---

## Nits

- Indentation in the SQL block uses tabs mixed with spaces in places — `gofmt` won't catch it inside a backtick string but it's slightly harder to scan than the surrounding queries.

edshav (approved):
Thanks for the fast turn-around — all seven items are addressed:

| # | Item | Resolution |
|---|---|---|
| 1 | max-of-day vs last-of-day | `e52f442d` reverted to last-of-day with explicit docstring |
| 2 | test for new semantics | `cb2fc88f` + `e52f442d` added an intraday-burn fixture with exact-value assertions |
| 3 | outer subquery cost | `e52f442d` switched to `blocks` table |
| 4 | NULL `script_type` | `e52f442d` uses `IS DISTINCT FROM` |
| 5 | issuance-input invariant | `e52f442d` added an explanatory SQL comment |
| 6 | misleading `total` alias | `e52f442d` renamed to `delta` throughout |
| 7 | reorg drift | `cb2fc88f` added `SKASupplyHeight` + reload-if-stale check |

Three follow-up observations on `cb2fc88f`, none blocking:

**A. Reload logic doesn't fall through when `GetHeight` errors and the cache is empty.**
`cmd/dcrdata/internal/api/apiroutes.go:1963`

```go
currHeight, err := c.DataSource.GetHeight(r.Context())
reload := false
if err != nil {
    log.Warnf("ChartTypeData: failed to get current height: %v", err)
} else {
    if cachedHeight, ok := c.charts.SKASupplyHeight(coinType); !ok || currHeight-cachedHeight > 10 {
        reload = true
    }
}
```

If `GetHeight` returns an error, `reload` stays `false` even if `c.charts.SKASupply[coinType]` is unpopulated. The subsequent `c.charts.Chart(...)` call will then return "no SKA supply data found" and the request 503s. Pre-change behavior at least attempted a load when cache was empty. Suggested seed:

```go
reload := !c.charts.SKASupplyExists(coinType)
if currHeight, err := c.DataSource.GetHeight(r.Context()); err == nil {
    if cachedHeight, ok := c.charts.SKASupplyHeight(coinType); !ok || currHeight-cachedHeight > 10 {
        reload = true
    }
} else {
    log.Warnf("ChartTypeData: failed to get current height: %v", err)
}
```

**B. `currHeight - cachedHeight > 10` is a magic number.**
Worth a named constant (e.g. `skaSupplyStaleHeightThreshold = 10`) with a one-line comment explaining the staleness budget (≈ N minutes given target block time). Otherwise the next reader has to infer why 10.

**C. The reload path writes to `charts.SKASupply[coinType]` without holding `charts.mtx`.**
`db/dcrpg/pgblockchain.go:1130`

`LoadSKASupplyForCoin` mutates the map directly, while `SKASupplyExists` and `SKASupplyHeight` read it under `charts.mtx.RLock()`. This is a pre-existing race, but the new "reload when stale" path means two concurrent chart requests for the same coin can now both decide to reload and race on the write. Three options: (1) take `charts.mtx.Lock()` around the map assignment in `LoadSKASupplyForCoin`, (2) add a per-coinType singleflight so concurrent requests dedupe, (3) leave it for a follow-up if you'd rather not expand scope.

Approval still stands.

PR #297, opened 2026-05-22
pubsub,cmd/dcrdata: migrate WebSocket to coder/websocket with RFC 6455 ping/pong

edshav:
## Summary

Replaces the deprecated `golang.org/x/net/websocket` library with `github.com/coder/websocket` v1.8.14 across both server hubs ([pubsub/pubsubhub.go](pubsub/pubsubhub.go), [cmd/dcrdata/internal/explorer/websockethandlers.go](cmd/dcrdata/internal/explorer/websockethandlers.go)) and the [psclient](pubsub/psclient/client.go) library, and adds **RFC 6455 ping/pong keepalive** so the server can detect and close zombie connections.

The motivation is that `x/net/websocket` has no RFC 6455 control-frame API, so the server cannot tell when a backgrounded client's TCP connection has silently died — the connection keeps occupying a hub slot, broadcast frames pile up in it, and the user sees a green "online" indicator while the table is silently frozen. After this change, each per-connection handler runs a ping goroutine that issues `ws.Ping(ctx)` every `pingInterval` and closes the socket on missed pong, so the client gets an honest disconnect signal.

### What changed

- **Both WebSocket handlers** now use `websocket.Accept` instead of `websocket.Handler{}.ServeHTTP`. Each connection runs three goroutines: receive, send, and a new **ping goroutine** that issues `ws.Ping(ctx)` every `pingInterval` and cancels the per-connection context on missed pong. A cancelled context tears down the other two goroutines and closes the socket.
- All `Set{Read,Write}Deadline` calls replaced with `context.WithTimeout`-bounded `wsjson.Read` / `wsjson.Write`. Per-read polling in `pubsubhub.receiveLoop` is dropped — liveness is now solely the ping goroutine's job.
- Writes from the receive and send loops rely on `coder/websocket`'s internal write serialization rather than an application-level mutex.
- **`pubsub/types`:** `IsWSClosedErr` and `IsIOTimeoutErr` rewritten in terms of `errors.Is(net.ErrClosed / io.EOF / context.Canceled)`, `websocket.CloseStatus(err) != -1`, and `context.DeadlineExceeded` — replacing the brittle substring match against `"use of closed network connection"` which only worked under `x/net/websocket`.
- **`psclient`:** drops the embedded `*websocket.Conn` from `Client`. In-tree [pubsub/democlient](pubsub/democlient/) compiles unchanged. `NewFromConn`'s parameter type changes to coder's `*websocket.Conn` — a breaking change for any external consumer that touches embedded `Conn` methods.
- Origin checking is preserved as open (`OriginPatterns: []string{"*"}`) on both handlers — the block explorer is a public read-only feed; tightening this is a separate decision.

### New tests

[pubsub/pubsubhub_ws_test.go](pubsub/pubsubhub_ws_test.go) — `httptest`-based round-trips:
- `TestWebSocketHandler_SubscribeRoundTrip` — subscribe / subscribeResp JSON shape compatibility.
- `TestWebSocketHandler_VersionRoundTrip` — version handshake (the path psclient hits immediately on dial).
- `TestWebSocketHandler_ConcurrentRequestsDoNotPanic` — multiple in-flight requests over one connection.

## Test plan

- [x] `go test ./...` passes in all 10 modules.
- [x] `golangci-lint` clean across all modules. The 2 leftover hits in [cmd/dcrdata/node_modules/flatted/golang/pkg/flatted/flatted.go](cmd/dcrdata/node_modules/flatted/golang/pkg/flatted/flatted.go) are pre-existing on `develop` (third-party npm-shipped Go code, unrelated to this PR — verified by stashing changes and re-linting on develop).
- [x] `go mod tidy` is a no-op in root, `cmd/dcrdata`, `db/dcrpg`, and `pubsub/democlient` after the change.
- [x] Pre-commit hook (`gofmt` + per-module `go test ./...`) green.
- [x] Playwright smoke against a locally running binary: home/blocks/mempool render, console clean, connection indicator stays green through `2 × pingInterval` (120s) — confirms the new keepalive works end-to-end with no spurious closes.
- [ ] **Reviewer: optional manual zombie test** — open the home page in Chrome, `kill -STOP <chrome_pid>` for ~70s, then `kill -CONT`. The server log should show a `WebSocketHandler` exit during the stop window (server closed the connection on missed pong), and the indicator should be red in the unfrozen tab.


yanchenko-igor (approved):
LGTM

PR #314, opened 2026-05-23
Fix/block fees pagination ska

yanchenko-igor:
Summary
This PR resolves critical bugs and inconsistencies in block fee accounting, SKA transaction reporting, and explorer pagination, while ensuring data parity between HTTP and WebSocket transports.
Key Changes
- Block Fee Accounting:
  - Updated block header totals to exclude stake-type transactions (TxTypeVote, TxTypeSSFee), ensuring the "Fee" total only reflects transaction and ticket fees.
  - Updated mining fee calculation in blockdata.go to subtract vote fees.
- SKA Transaction Reporting:
  - Fixed a field shadowing bug in TrimmedTxInfo where CoinType was being read from the outer struct instead of the embedded TxBasic, causing SKA transactions to be misidentified as VAR.
  - Implemented high-precision fee and fee-rate calculations for SKA transactions using big.Int to prevent float64 precision loss.
  - Updated the block view template to render these high-precision fees using the correct coin symbols (e.g., SKA1/B).
- Pagination Logic:
  - Fixed a bug in calcPages and calcPagesDesc that caused non-existent trailing pages to be displayed when the total item count was an exact multiple of the page size.
- Testing & Verification:
  - Added comprehensive internal tests in db/dcrpg/pgblockchain_internal_test.go to verify the trimmedTxInfoFromMsgTx logic for VAR, SKA, and Coinbase transactions.
  - Verified fixes via unit tests for pagination.
Relevant Files
- db/dcrpg/pgblockchain.go: Core fee aggregation and transaction transformation logic.
- db/dcrpg/pgblockchain_internal_test.go: New test suite for transaction fee processing.
- blockdata/blockdata.go: Mining fee calculation logic.
- cmd/dcrdata/internal/explorer/explorerroutes.go: Pagination calculation logic.
- cmd/dcrdata/views/block.tmpl: HTML template for block fee rendering.

### DIFF
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -667,5 +667,14 @@ func computeMiningFee(msgBlock *wire.MsgBlock, powSubsidy int64) int64 {
 	}
 	miningFee := coinbaseOutput - powSubsidy
 
-	return miningFee
+	// Exclude stake fees (fees from vote transactions) from the mining fee
+	var voteFees int64
+	for _, tx := range msgBlock.Transactions {
+		if txhelpers.DetermineTxType(tx) == stake.TxTypeSSGen {
+			fee, _ := txhelpers.TxFeeRate(tx)
+			voteFees += int64(fee)
+		}
+	}
+
+	return miningFee - voteFees
 }
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -247,7 +247,7 @@ func (t *Collector) CollectBlockInfo(hash *chainhash.Hash) (*apitypes.BlockDataB
 	}
 	var miningFee int64
 	if curSubsidy != nil {
-		miningFee = computeMiningFee(msgBlock, curSubsidy.PoW)
+		miningFee = computeMiningFee(msgBlock)
 	}
 
 	// Work/Stake difficulty
@@ -655,26 +655,21 @@ func BlockSKAPoWRewardsFromSTx(msgBlock *wire.MsgBlock) map[uint8]string {
 	return out
 }
 
-// computeMiningFee calculates total mining fees from a block.
-// Formula: Mining Fee = Coinbase Output - PoW Subsidy
-// This is equivalent to summing (Inputs - Outputs) for all non-coinbase transactions,
-// but more efficient and accurate.
-func computeMiningFee(msgBlock *wire.MsgBlock, powSubsidy int64) int64 {
-	// Get coinbase output (first transaction, first output)
-	coinbaseOutput := int64(0)
-	for _, txOut := range msgBlock.Transactions[0].TxOut {
-		coinbaseOutput += txOut.Value
-	}
-	miningFee := coinbaseOutput - powSubsidy
-
-	// Exclude stake fees (fees from vote transactions) from the mining fee
-	var voteFees int64
-	for _, tx := range msgBlock.Transactions {
-		if txhelpers.DetermineTxType(tx) == stake.TxTypeSSGen {
-			fee, _ := txhelpers.TxFeeRate(tx)
-			voteFees += int64(fee)
+// computeMiningFee calculates total mining fees from a block by summing fees
+// from regular transactions and tickets, excluding votes and stake fees (SSFee).
+// For Monetarium, fees included in MiningFee are only from Tx and Tickets.
+func computeMiningFee(msgBlock *wire.MsgBlock) int64 {
+	var miningFee int64
+	for i, tx := range msgBlock.Transactions {
+		if i == 0 {
+			continue // skip coinbase
 		}
+		txType := txhelpers.DetermineTxType(tx)
+		if txType == stake.TxTypeSSGen || txType == stake.TxTypeSSFee {
+			continue // exclude votes and stake fees
+		}
+		fee, _ := txhelpers.TxFeeRate(tx)
+		miningFee += int64(fee)
 	}
-
-	return miningFee - voteFees
+	return miningFee
 }
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -656,19 +656,24 @@ func BlockSKAPoWRewardsFromSTx(msgBlock *wire.MsgBlock) map[uint8]string {
 }
 
 // computeMiningFee calculates total mining fees from a block by summing fees
-// from regular transactions and tickets, excluding votes and stake fees (SSFee).
-// For Monetarium, fees included in MiningFee are only from Tx and Tickets.
+// from regular transactions (Transactions tree) and tickets (STransactions tree
+// limited to TxTypeSStx), matching the GetExplorerBlock formula:
+//
+//	getTotalFee(block.Tx) + getTotalFee(block.Tickets)
 func computeMiningFee(msgBlock *wire.MsgBlock) int64 {
 	var miningFee int64
 	for i, tx := range msgBlock.Transactions {
 		if i == 0 {
 			continue // skip coinbase
 		}
-		txType := txhelpers.DetermineTxType(tx)
-		if txType == stake.TxTypeSSGen || txType == stake.TxTypeSSFee {
-			continue // exclude votes and stake fees
-		}
 		fee, _ := txhelpers.TxFeeRate(tx)
+		miningFee += int64(fee)
+	}
+	for _, stx := range msgBlock.STransactions {
+		if txhelpers.DetermineTxType(stx) != stake.TxTypeSStx {
+			continue // only include tickets
+		}
+		fee, _ := txhelpers.TxFeeRate(stx)
 		miningFee += int64(fee)
 	}
 	return miningFee
--- a/blockdata/blockdata_test.go
+++ b/blockdata/blockdata_test.go
@@ -409,3 +409,7 @@ func TestComputeMiningFee_MultipleRegularTx(t *testing.T) {
 		t.Errorf("got %d, want %d", got, want)
 	}
 }
+
+func TestComputeMiningFee_STransactions(t *testing.T) {
+	t.Skip("Skipping: constructing valid SStx/SSGen transactions requires specific OP_RETURN structure")
+}
--- a/blockdata/blockdata_test.go
+++ b/blockdata/blockdata_test.go
@@ -4,6 +4,7 @@ import (
 	"math/big"
 	"testing"
 
+	"github.com/monetarium/monetarium-node/blockchain/stake"
 	"github.com/monetarium/monetarium-node/cointype"
 	"github.com/monetarium/monetarium-node/wire"
 )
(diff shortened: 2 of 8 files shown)


edshav (commented):
## TL;DR

Mix of a clearly-correct bug fix (CoinType field shadowing), a needed pagination fix that's separate from #312, a reasonable block-fee accounting change, and a `FeeRateRaw` unit change that creates inconsistency with how the same field is computed elsewhere. Issue #303's parity-test gate isn't addressed.

## On the pagination overlap with #312

The pagination changes here are **still needed** after #312 and fix a different bug:
- #312 corrected the *source* of `rows` for `/days`, `/weeks` via `COUNT(DISTINCT DATE_TRUNC(...))`. It did not touch `calcPages` / `calcPagesDesc`.
- This PR fixes a real off-by-one in those functions: the old `pages := rows/pageSize + 1` overshoots whenever `rows` is an exact multiple of `pageSize`, and returns empty for `rows < pageSize`. The new ceiling-division formula is correct.

Examples where #312's clean row count still rendered a ghost page before this PR:
- `rows=20, pageSize=10` → old: 3 pages, correct: 2.
- `rows=10, pageSize=10` → old: 2 pages, correct: 1.
- `rows=6, pageSize=10` → old: empty, correct: 1 (matches the new `expected: "1"` test assertion).

Keep the pagination changes.

## Findings on the rest

### 1. CoinType shadowing fix — correct ✅
`TrimmedTxInfo.CoinType` (outer) was shadowing `TxBasic.CoinType` (inner) and was never set, so templates always rendered VAR. Setting `CoinType: txBasic.CoinType` at construction fixes this. The explicit overrides at `db/dcrpg/pgblockchain.go:6731-6734` (SSFee) and `:6751-6753` (regular tx) are now redundant — worth a one-line cleanup but not blocking.

### 2. Block header "Fee" exclusions — match issue #303, but go wider than asked
- Per-tx skip of `TxTypeVote` / `TxTypeSSFee` at `pgblockchain.go:6810-6818` lines up with #303.
- However the `MiningFee` aggregate at `:6839` now excludes `block.Revs` and `block.Treasury` too (was: `Tx + Treasury + Revs + Tickets + Votes`; now: `Tx + Tickets`). Issue #303 only mentions excluding vote / stake-fee txs. Is the wider exclusion intentional? Please confirm explicitly.

### 3. `computeMiningFee` double-subtraction concern
The original comment said the function is "equivalent to summing (Inputs − Outputs) for all non-coinbase transactions". If true, vote fees were already excluded and the new explicit subtraction is double-counting. The comment was likely wrong (stakebase folded into `coinbaseOutput`), but worth confirming with a numeric trace on a real block that has SSGen fees, not just unit tests on synthetic msgTxs.

### 4. `FeeRateRaw` unit-of-measure inconsistency ⚠️
The same struct field is now populated with two different units across three constructors in `db/dcrpg/pgblockchain.go`:
- `GetAPITransaction` (SKA), `:5542-5550`: `fee * 1000 / size` = **atoms/kB**.
- `GetExplorerTx` (SKA), `:7088-7095`: same — **atoms/kB**.
- `trimmedTxInfoFromMsgTx` (VAR + SKA, this PR), `:6594-6605`: `fee / size` = **atoms/byte**.

Consumers also disagree: `block.tmpl:419` (this PR) treats it as atoms/byte and renders `… {{coinSymbol .CoinType}}/B`; `tx.tmpl:243` treats it as atoms/kB and renders `… {{coinSymbol .CoinType}}/kB`. A future reader can't tell which unit a given instance carries. Please either standardize on one unit across all three constructors, or document the unit on the `FeeRateRaw` field with a comment at each construction site.

Secondary issues from the same change:
- For VAR, the block-table column changes from `dcrPerKbToAtomsPerByte .FeeRate` (e.g. `50 atoms/B`, integer) to `coinDecimalParts .FeeRateRaw .CoinType` (e.g. `0.00000050 VAR/B`). Functionally equivalent, but harder to scan. Worth a UX confirmation with a screenshot.
- Precision: `rate := fee / txSize` truncates to zero whenever `fee < txSize` (e.g. fee=100, size=500 → "0"). The old SKA path multiplied by 1000 first to avoid this. The new atoms/byte unit can't represent sub-atom-per-byte fee rates at all.

### 5. Issue #303's "both-transports parity" gate is not addressed ⚠️
#303 says explicitly: *"identical field set and identical values for the Fee total, Transactions[].FeeRate, and StakeFees[].Fee on the HTTP template context and the WebSocket JSON for the same block. This is the non-negotiable gate per the area's wiki notes."*

The new test in `pgblockchain_internal_test.go` only covers the HTTP side (`trimmedTxInfoFromMsgTx` on isolated msgTxs). No WebSocket-side change in `pubsub/pubsubhub.go`, no parity test. Either the parity already holds and a test should lock it in, or there's an open gap on the pubsub path. Please address.

#303 also asks for unit tests covering blocks with "only VAR txs, only SKA txs, and mixed VAR+SKA{n} blocks". The new test covers a single VAR, a single SKA, and a coinbase tx in isolation — not block-level scenarios.

### 6. Stray indentation in `GetMempool` — cosmetic
`pgblockchain.go:7494` changes `txType := txhelpers.DetermineTxType(msgTx)` to one tab (was two) and adds blank lines. The whole `GetMempool` body is inside a `/* … */` block comment marked "UNUSED?", so `gofmt` doesn't catch it and runtime is unaffected — but the change looks accidental. Worth reverting that hunk.

### 7. Tests
- Pagination test updates (`expected: "1"` for `rows=6, pageSize=10`) are correct and good — they lock in the previously broken edge case.
- The new fee-calculation test uses synthetic `wire.MsgTx` literals; assertions hold but value is limited. A table-driven test against a real serialized block, comparing `FeeRaw`/`FeeRateRaw` *and* the pubsub JSON for the same block, would directly answer #303.

## Summary of asks

1. Decide and document the `FeeRateRaw` unit contract — ideally unify atoms/byte vs atoms/kB across all three constructors and both templates.
2. Add an HTTP ↔ WebSocket parity test for the block view-model (required by #303).
3. Confirm whether dropping `block.Revs` and `block.Treasury` from `block.MiningFee` is intentional, since #303 only asks for vote / SSFee exclusion.
4. Revert the stray whitespace change in the commented-out `GetMempool`.
5. Verify `computeMiningFee` against a real block with SSGen fees to confirm there's no double-subtraction.

edshav (commented):
Corrections to my previous review — three items I got wrong on a second pass:

### Correction 1 — Finding #2 was wrong

I claimed the `block.MiningFee` aggregate going from `Tx + Treasury + Revs + Tickets + Votes` down to `Tx + Tickets` "goes wider than asked" by issue #303. That's incorrect. Issue #303 explicitly says:

> Header "Fee" total sums fees **only** from the Transactions and Tickets tables — exclude the Stake Fees table from this total.

`Tx + Tickets` is exactly that. ✅ The change matches the issue. Disregard that point.

### Correction 2 — Finding #3 is a real bug, not a "sanity check"

The new vote-fee subtraction in `computeMiningFee` at [blockdata/blockdata.go:670-679](blockdata/blockdata.go#L670-L679) iterates the wrong slice:

```go
for _, tx := range msgBlock.Transactions {
    if txhelpers.DetermineTxType(tx) == stake.TxTypeSSGen { ... }
}
```

SSGen (vote) transactions live in `msgBlock.STransactions`, not `msgBlock.Transactions`. Existing code in the same file confirms this — see [blockdata.go:440](blockdata/blockdata.go#L440) (`allTxs := append(msgBlock.Transactions, msgBlock.STransactions...)`) and [blockdata.go:443](blockdata/blockdata.go#L443) which filters SSGen out of an "all txs" slice. So `voteFees` will always be 0 and the subtraction has no effect. The fix as written doesn't accomplish what the PR description claims ("Updated mining fee calculation in blockdata.go to subtract vote fees"). Iterate `msgBlock.STransactions` (or both trees) instead.

### Correction 3 — Line numbers in the earlier review

Several `:line` refs in my earlier comment cited pre-merge diff context lines instead of the file lines on this branch. The substantive findings stand; the correct anchors are:

- `FeeRateRaw` units (the inconsistency claim is correct):
  - **atoms/kB**, `GetAPITransaction` SKA path: [db/dcrpg/pgblockchain.go:5529-5537](db/dcrpg/pgblockchain.go#L5529-L5537) (I said 5542-5550)
  - **atoms/kB**, `GetExplorerTx` SKA path: [db/dcrpg/pgblockchain.go:7083-7090](db/dcrpg/pgblockchain.go#L7083-L7090) (I said 7088-7095)
  - **atoms/byte**, `trimmedTxInfoFromMsgTx` (this PR): [db/dcrpg/pgblockchain.go:6569-6602](db/dcrpg/pgblockchain.go#L6569-L6602) (I said 6594-6605)
- Redundant CoinType overrides now that `trimmedTxInfoFromMsgTx` sets it:
  - SSFee: [db/dcrpg/pgblockchain.go:6755-6760](db/dcrpg/pgblockchain.go#L6755-L6760) (I said 6731-6734)
  - Regular tx: [db/dcrpg/pgblockchain.go:6775-6778](db/dcrpg/pgblockchain.go#L6775-L6778) (I said 6751-6753)
- Per-tx exclusion of Vote/SSFee in the fee-total loop: [db/dcrpg/pgblockchain.go:6816-6819](db/dcrpg/pgblockchain.go#L6816-L6819)
- `block.MiningFee` aggregate at [db/dcrpg/pgblockchain.go:6839](db/dcrpg/pgblockchain.go#L6839) — line number was correct.
- Stray indentation at [db/dcrpg/pgblockchain.go:7494](db/dcrpg/pgblockchain.go#L7494) — correct.

### Net updated ask list

1. ✅ Pagination changes — keep, separate from #312.
2. ✅ CoinType shadowing fix — correct.
3. ✅ `MiningFee = Tx + Tickets` — matches #303 (retracted my earlier objection).
4. ❌ `computeMiningFee` vote-fee subtraction — **iterate `STransactions` (or both trees)**, the current loop is a no-op.
5. ⚠️ `FeeRateRaw` unit inconsistency (atoms/byte vs atoms/kB across three constructors and two templates) — decide & document the contract.
6. ⚠️ #303's HTTP↔WebSocket parity test — still not addressed.
7. nit: revert stray whitespace at line 7494 (commented-out `GetMempool`).

edshav (commented):
Re-reviewed bb73fd1.

### Resolved ✅

- **FeeRateRaw unit standardization**: all three constructors now produce atoms/kB.
  - VAR path uses `feeRate` (2nd return from `txhelpers.TxFeeRate`, which is atoms/kB by its doc comment).
  - SKA path multiplies by 1000 before dividing by size.
  - `block.tmpl` updated to `/kB`. Field semantics now consistent with `tx.tmpl:243`, `GetAPITransaction`, and `GetExplorerTx`.
  - Test asserts the exact atoms/kB value. Good.
- **Stray indentation at `pgblockchain.go:7494`**: fixed.

### New concern with the `computeMiningFee` rewrite ⚠️

The new [`computeMiningFee`](blockdata/blockdata.go) iterates only `msgBlock.Transactions` (regular tree):

```go
for i, tx := range msgBlock.Transactions {
    if i == 0 { continue }
    txType := txhelpers.DetermineTxType(tx)
    if txType == stake.TxTypeSSGen || txType == stake.TxTypeSSFee {
        continue
    }
    fee, _ := txhelpers.TxFeeRate(tx)
    miningFee += int64(fee)
}
```

Two problems:

1. **Vote / SSFee filter is dead code.** `TxTypeSSGen` and `TxTypeSSFee` txs live in `msgBlock.STransactions`, not `msgBlock.Transactions`. The cross-tree convention in this file is visible at [blockdata.go:440](blockdata/blockdata.go#L440) (`allTxs := append(msgBlock.Transactions, msgBlock.STransactions...)`) and the stake-only iteration at [blockdata.go:615](blockdata/blockdata.go#L615). So the `txType ==` filter never fires.

2. **Tickets are not included despite the commit message saying so.** The commit message says *"sum fees from regular transactions **and tickets**, skipping coinbase, votes, SSFees"*. But tickets (`SStx`) also live in `STransactions` (see the switch at [pgblockchain.go:6727-6760](db/dcrpg/pgblockchain.go#L6727-L6760) where `RawSTx → TxTypeSStx → tickets`). So this function only sums **regular-tx fees**, not ticket fees.

   That's a real divergence from the explorer path: [`GetExplorerBlock`](db/dcrpg/pgblockchain.go) sets `block.MiningFee = (getTotalFee(block.Tx) + getTotalFee(block.Tickets)).ToCoin()` — it *does* include ticket fees. So `Collector.CollectBlockInfo` and `GetExplorerBlock` now compute different `MiningFee` values on any block that has ticket purchases with fees. The PR description claims this "better aligns with the GetExplorerBlock MiningFee calculation" — but it doesn't.

Suggested fix: iterate both `msgBlock.Transactions` (skip coinbase) and `msgBlock.STransactions` (keep only `TxTypeSStx`), or document the asymmetry explicitly. The current unit tests can't catch this because none of them populate `msgBlock.STransactions`. A test with a coinbase + one regular tx + one SStx would exercise the intended behavior.

### Still open

- **#303 HTTP ↔ WebSocket parity test**: still not addressed.

edshav (approved):
Approving after re-review of bb73fd1..b4020901.

The two `computeMiningFee` concerns from my prior comment are resolved by 2fae3d35:

- The dead `Vote/SSFee` filter is gone from the regular-tree loop, replaced by an `STransactions` loop whose `!= TxTypeSStx` filter actually fires on the correct slice.
- Tickets are now summed, matching `getTotalFee(block.Tx) + getTotalFee(block.Tickets)` from `db/dcrpg/pgblockchain.go:6836`. Votes / SSFee / SSRtx remain excluded.

Test coverage for the new `STransactions` code path is added in b4020901 (TicketInSTransactions, BothTrees, NonTicketSTransactionsExcluded) — uses an `IsSStx`-valid ticket constructed from `stake`-package fixture scripts.

Carried forward as separate work:

- The `#303` HTTP ↔ WebSocket parity test is still not in this branch.

edshav (commented):
Re-reviewed `bda9135200`, which addresses the only carried-forward item from my prior approval (#303 HTTP↔WS parity test). The new `BlockFeeParity` subtest passes locally and probes the right fields, but it's strictly a WS round-trip check rather than a parity check.

### Concrete issues

1. **No HTTP wire is generated.** The sibling test [`BlockWSWireFormatEquivalence`](cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L87-L119) already established the exact pattern this should follow: HTTP wire = `bi.Trim(...) → json.Marshal`, WS wire = `blockCopy → json.Marshal`, then cross-compare. The new test only marshals the WS side. All comparisons are WS-decoded-value vs. the input struct field:
   - [Line 165](cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L165): `wsMiningFee != feeBI.MiningFee` is tautologically true unless `json.Marshal` corrupts a `float64`. The inline comment "HTTP template reads .MiningFee from the same BlockInfo struct" argues *why* parity must hold, but #303 asks for an explicit parity assertion across transports — not the argument.

2. **`Tx[]` assertion is non-empty, not equal.** [Lines 180-187](cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L180-L187) only check `feeRaw == ""` / `feeRateRaw == ""`. The `StakeFees[]` loop immediately below ([lines 199-205](cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L199-L205)) correctly asserts `feeRaw != feeBI.StakeFees[i].FeeRaw`. `Tx[]` should match that pattern — otherwise a regression that emits wrong digits for regular-tx fee fields slips through.

3. **`Trim()` is bypassed.** Production WS encodes `blockCopy` *after* `block.Trim(maxBlockSize, issuedSKA)` and the three-field patch at [websockethandlers.go:285-288](cmd/dcrdata/internal/explorer/websockethandlers.go#L285-L288). The new test marshals `feeBI` directly. Fee fields aren't touched by `Trim` today, so the test happens to pass — but a future change to `Trim` that affected (or shadowed) fee fields wouldn't be caught.

4. **Fixture values don't lock in the unit contract.** The PR's headline outcome is `FeeRateRaw` in atoms/kB. The fixtures use atoms/kB-shaped numbers (`"500000"`, `"125000000000000000"`) but aren't derived from `feeAtoms * 1000 / size`, so a regression that flipped the unit back to atoms/byte would still pass.

### Asks

- Mirror `BlockWSWireFormatEquivalence`: marshal `bi.Trim(...)` as the HTTP wire, marshal the trimmed `blockCopy` as the WS wire, and compare the fee fields *between the two wires*.
- Change the `Tx[]` loop to assert `feeRaw == feeBI.Tx[i].FeeRaw` and the same for `FeeRateRaw`, skipping coinbase.

Both are mechanical and would actually realize the parity gate from #303. Given that this is test-only on top of an already-approved branch, I'd rather see the tightenings here than carry them as another open item.

edshav (approved):
Re-reviewed against the prior comment on `bda9135200`. Commit `08c8edef` ("tighten BlockFeeParity to true cross-wire test") addresses all four asks; approving.

**1. HTTP wire is now generated.** [visualblocks_contract_test.go#L139-L154](https://github.com/monetarium/monetarium-explorer/blob/08c8edef4b2a270ce42a84b06f9eaeab31d2df8b/cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L139-L154) calls `feeBI.Trim(...)` then marshals `trimmed` as the HTTP wire and the patched `blockCopy` as the WS wire — same shape as `BlockWSWireFormatEquivalence`.

**2. `Tx[]` is now true equality.** [Lines 213-218](https://github.com/monetarium/monetarium-explorer/blob/08c8edef4b2a270ce42a84b06f9eaeab31d2df8b/cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L213-L218) compare `wTx["FeeRaw"]` to `hTx["FeeRaw"]` (same for `FeeRateRaw`), matching the `StakeFees[]` pattern.

**3. `Trim()` is no longer bypassed.** The WS-side `blockCopy` patches `CoinFills / ActiveSKACount / MaxBlockSize / RegularCoinCounts` from `trimmed`, mirroring production at [websockethandlers.go#L284-L289](https://github.com/monetarium/monetarium-explorer/blob/08c8edef4b2a270ce42a84b06f9eaeab31d2df8b/cmd/dcrdata/internal/explorer/websockethandlers.go#L284-L289).

**4. Fixtures lock in atoms/kB.** [Lines 128-138](https://github.com/monetarium/monetarium-explorer/blob/08c8edef4b2a270ce42a84b06f9eaeab31d2df8b/cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L128-L138) derive constants as `fee * 1000 / size` (`10000000 * 1000 / 500 = 20000000`, etc.) and [lines 220-225](https://github.com/monetarium/monetarium-explorer/blob/08c8edef4b2a270ce42a84b06f9eaeab31d2df8b/cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L220-L225) explicitly assert the HTTP wire equals the derived `feeRateRaw`. An atoms/byte regression would now fail this test.

---

Nit (non-blocking): the `blockCopy` patch in `BlockFeeParity` covers 4 fields, but production [websockethandlers.go#L285-L289](https://github.com/monetarium/monetarium-explorer/blob/08c8edef4b2a270ce42a84b06f9eaeab31d2df8b/cmd/dcrdata/internal/explorer/websockethandlers.go#L285-L289) and the sibling [`BlockContractWireFormat` subtest at lines 100-104](https://github.com/monetarium/monetarium-explorer/blob/08c8edef4b2a270ce42a84b06f9eaeab31d2df8b/cmd/dcrdata/internal/explorer/visualblocks_contract_test.go#L100-L104) also patch `TotalFillRatio`. Fee assertions don't touch it so nothing fails today; worth adding `blockCopy.TotalFillRatio = trimmed.TotalFillRatio` for consistency with the production patch list.

edshav (approved):

PR #315, opened 2026-05-25
Fix/address stake metrics

yanchenko-igor:
PR Summary: Fix Address Balance Totals and Stake Ratios
Description
This PR corrects the calculation of balance totals and "Stake income" / "Stake spending" ratios on the address page. It addresses issues where burned coins were included in totals, reward attribution was too coarse, and ratios were skewed by system-level issuance (PoW) and change outputs. Additionally, it resolves a critical performance bottleneck that caused timeouts for addresses with high transaction volumes.
Key Changes
1. Balance & Total Calculations
- Burn Exclusion: Updated the address balance query to exclude outputs with script_type = 'nulldata', ensuring burned coins do not inflate "Balance" or "Received" totals.
- Accurate Spending: Fixed a bug where TotalSpent for VAR was derived as TotalReceived - TotalUnspent. It now uses the actual summed value of spending transactions from the database.
2. Stake Ratio Refinements
- Granular Reward Attribution: Introduced new transaction types (TxTypeBlockRewardPoS, TxTypeBlockRewardPoW, TxTypeSSFeePoS, TxTypeSSFeePoW, TxTypeTicketPurchase) to distinguish between PoS and PoW income.
- Income Ratio: The "Stake income %" denominator now correctly excludes PoW coinbase rewards and change outputs, while the numerator includes PoS rewards and SS fees.
- Spending Ratio: Implemented an explicit allow-list for stake spending (Ticket Purchases, PoS SS Fees, SSGen, and SSRtx), preventing regular spends from being misattributed as stake spending.
3. Extraction & Persistence
- Ticket Classification: Improved the extraction pipeline to identify ticket purchases using script inspection (SCStakeSubmission / SCStakeSubChange). This prevents the explorer from misclassifying ticket purchases as "Regular" transactions when the node's general classification is ambiguous.
- Persistence Update: Updated the UpsertAddressRow logic to ensure tx_type is updated on conflict, allowing the database to reflect corrected classifications.
4. Performance Optimization
- Timeout Resolution: Rewrote the SelectAddressSpentUnspentCountAndValue query to eliminate a costly EXISTS subquery used for change detection. By replacing it with a Common Table Expression (CTE) and a LEFT JOIN, query times for large addresses (~32k+ rows) were reduced from 20+ minutes to milliseconds.
Verification Results
- Ratio Accuracy: Verified that addresses with 100% stake income/spending are correctly displayed.
- Performance: Confirmed that large addresses no longer trigger database timeouts.
- Consistency: Verified that $\text{Balance} = \text{Received} - \text{Spent}$ holds across all coin types.

### DIFF
--- a/cmd/dcrdata/views/address.tmpl
+++ b/cmd/dcrdata/views/address.tmpl
@@ -126,16 +126,33 @@
                   {{- end}}
                 {{- end}}
               {{- end}}
-              {{- if and $varCB $balance.HasStakeOutputs}}
-                <div class="col-12 pb-2 fs14 text-secondary text-start">
-                    <span class="fw-bold">Stake spending</span>: {{printf "%.1f" (x100 $balance.FromStake)}}%
-                </div>
-              {{- end}}
-              {{- if and $varCB $balance.HasStakeInputs}}
-                <div class="col-12 pb-2 fs14 text-secondary text-start">
-                    <span class="fw-bold">Stake income</span>: {{printf "%.1f" (x100 $balance.ToStake)}}%
-                </div>
-              {{- end}}
+                {{- if .Balance.HasStakeOutputs}}
+                    <div class="col-12 pb-2 fs14 text-secondary text-start">
+                        <span class="fw-bold">Stake income</span>:
+                        {{- range $ct := .ActiveCoins}}
+                          {{- $cb := index $balance.Coins $ct}}
+                          {{- if $cb}}
+                          {{- if gt $cb.FromStake 0.0}}
+                            {{printf "%.1f" (x100 $cb.FromStake)}}% {{coinSymbol $ct}}<br>
+                          {{- end}}
+                        {{- end}}
+                      {{- end}}
+                  </div>
+                {{- end}}
+                {{- if .Balance.HasStakeInputs}}
+                    <div class="col-12 pb-2 fs14 text-secondary text-start">
+                        <span class="fw-bold">Stake spending</span>:
+                        {{- range $ct := .ActiveCoins}}
+                          {{- $cb := index $balance.Coins $ct}}
+                          {{- if $cb}}
+                          {{- if gt $cb.ToStake 0.0}}
+                            {{printf "%.1f" (x100 $cb.ToStake)}}% {{coinSymbol $ct}}<br>
+                          {{- end}}
+                        {{- end}}
+                      {{- end}}
+                  </div>
+                {{- end}}
+
           </div>
           {{/* <div class="row pb-3 fs16">
             <span class="col-24"><a href="{{$.Links.DownloadLink}}" title="Decred downloads" target="_blank" rel="noopener noreferrer">Get Decrediton</a>, the official desktop wallet.</span>
--- a/cmd/dcrdata/views/address.tmpl
+++ b/cmd/dcrdata/views/address.tmpl
@@ -126,32 +126,32 @@
                   {{- end}}
                 {{- end}}
               {{- end}}
-                {{- if .Balance.HasStakeOutputs}}
-                    <div class="col-12 pb-2 fs14 text-secondary text-start">
-                        <span class="fw-bold">Stake income</span>:
-                        {{- range $ct := .ActiveCoins}}
-                          {{- $cb := index $balance.Coins $ct}}
-                          {{- if $cb}}
-                          {{- if gt $cb.FromStake 0.0}}
-                            {{printf "%.1f" (x100 $cb.FromStake)}}% {{coinSymbol $ct}}<br>
-                          {{- end}}
-                        {{- end}}
-                      {{- end}}
-                  </div>
+            {{- if .Balance.HasStakeOutputs}}
+              <div class="col-12 pb-2 fs14 text-secondary text-start">
+                <span class="fw-bold">Stake income</span>:
+                {{- range $ct := .ActiveCoins}}
+                  {{- $cb := index $balance.Coins $ct}}
+                  {{- if $cb}}
+                    {{- if gt $cb.FromStake 0.0}}
+                      {{printf "%.1f" (x100 $cb.FromStake)}}% {{coinSymbol $ct}}<br>
+                    {{- end}}
+                  {{- end}}
                 {{- end}}
-                {{- if .Balance.HasStakeInputs}}
-                    <div class="col-12 pb-2 fs14 text-secondary text-start">
-                        <span class="fw-bold">Stake spending</span>:
-                        {{- range $ct := .ActiveCoins}}
-                          {{- $cb := index $balance.Coins $ct}}
-                          {{- if $cb}}
-                          {{- if gt $cb.ToStake 0.0}}
-                            {{printf "%.1f" (x100 $cb.ToStake)}}% {{coinSymbol $ct}}<br>
-                          {{- end}}
-                        {{- end}}
-                      {{- end}}
-                  </div>
+              </div>
+            {{- end}}
+            {{- if .Balance.HasStakeInputs}}
+              <div class="col-12 pb-2 fs14 text-secondary text-start">
+                <span class="fw-bold">Stake spending</span>:
+                {{- range $ct := .ActiveCoins}}
+                  {{- $cb := index $balance.Coins $ct}}
+                  {{- if $cb}}
+                    {{- if gt $cb.ToStake 0.0}}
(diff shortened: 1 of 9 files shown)


edshav (changes requested):
Nice fix — the label swap correction alone is worth merging, and the granular tx_type model is the right shape for multi-coin reward attribution. A few items before this is mergeable:

## Must fix

**1. Debug print left in `db/dcrpg/queries.go:1576`**
```go
fmt.Printf("DEBUG: Skipping spent row: matching_tx_hash is NULL for address %s\n", address)
```
The existing `log.Errorf` below it already covers this — drop the `Printf`.

**2. Magic numbers in stake-spending classifier** (`queries.go`, new branch)
```go
if txType == dbtypes.TxTypeTicketPurchase || txType == dbtypes.TxTypeSSFeePoS || txType == 2 || txType == 3 {
```
`2`/`3` are `stake.TxTypeSSGen`/`SSRtx`. Other call sites use `int16(stake.TxTypeSSGen)` (e.g. `pgblockchain.go:4764`) — please use the constants; an upstream enum reorder would silently break this.

**3. Dead empty block** (`pgblockchain.go:2476`)
```go
if balance != nil && balance.Coins != nil {
}
```
Leftover from removing the flat-VAR sync — delete it.

**4. Dead accumulator `skaSpent`**
`skaSpent` is populated in the spending branch but never read; the final loop computes `TotalSpentSKA = received - unspent`. Either remove it, or use it (and add a comment on why the SKA path derives spent from `received - unspent` while VAR uses the actual sum — they're inconsistent right now and the PR description specifically calls out VAR's switch to the actual sum).

## Worth confirming

**5. Possible `FromStake > 1.0`**
The CTE marks any funding row as `is_change` if the address also has a spending row in the same tx. Numerator/denominator are then decoupled:
```go
if txType != dbtypes.TxTypeBlockRewardPoW && !isChange {
    receivedExclIssuanceVAR += totalValue    // denominator
}
if txType == dbtypes.TxTypeBlockRewardPoS || txType == dbtypes.TxTypeSSFeePoS {
    fromStakeVAR += totalValue                // numerator
}
```
If any future or unusual income-type tx happens to also have a spend row for the same address (self-pay, certain SSFee constructions), the value lands in the numerator but not the denominator → ratio > 1. Either skip `isChange` exclusion when `txType` is an income type, or at least assert `FromStake <= 1.0` in the test.

**6. SSGen reward attribution**
`fromStakeVAR` counts `TxTypeBlockRewardPoS` (101) and `TxTypeSSFeePoS` (103). SSGen outputs (`stake.TxTypeSSGen = 2`) are not in the list. If on Monetarium stakers receive any portion of their payout via SSGen outputs (vs. purely via SSFee), this under-counts income. Could you confirm in the PR description that SSFee fully covers PoS payouts here?

**7. tx_type namespace overlap**
New constants `101..105` are cast to `stake.TxType` and stored in the same column as upstream values (`0..6`). Fine today, but if upstream ever grows to 101 there's a silent collision. A one-line comment on `types.go:51` noting "values ≥100 are dbtypes-only extensions to `stake.TxType`" would help, or move them to a clearly distinct range (e.g. `1000+`).

**8. Template** (`address.tmpl:126-150`)
Braces balance but the `{{- end}}` indentation is misleading (looks like 3 closes when there are 4) — please re-indent. Also note that the outer guard uses `HasStakeOutputs` (iterates `Coins`) while the inner range iterates `ActiveCoins`; if those ever disagree the div renders empty.

## Test coverage

The new `balance_test.go` exercises the **query** but builds state via raw SQL inserts, so the changes in `db/dbtypes/extraction.go` are untested:
- ticket-purchase reclassification via `SCStakeSubmission`/`SCStakeSubChange`
- SSFee marker → PoS/PoW split
- coinbase index-0 vs index>0 split

Even one extraction-path test plus a `FromStake <= 1.0` assertion in the existing test would close most of this.

## Things I'd love to see in the PR body

- An `EXPLAIN ANALYZE` excerpt confirming the 20+ min → ms claim (purely so reviewers can sanity-check the new plan is index-friendly).
- A line on whether SSGen payouts exist on Monetarium and how they're attributed (re #6).

edshav (approved):
### Follow-up review (commits e15efe10, 0e0bfb5a, 47263c52)

Thanks for the quick turnaround — the four must-fix items and three of four worth-confirming items are resolved:

- Debug `fmt.Printf` removed at [queries.go#L1556-L1559](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/queries.go#L1556-L1559).
- Magic `2`/`3` replaced with `int16(stake.TxTypeSSGen)`/`SSRtx` at [queries.go#L1584-L1588](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/queries.go#L1584-L1588).
- Dead empty `if` block removed at [pgblockchain.go#L2472-L2478](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/pgblockchain.go#L2472-L2478).
- `skaSpent` now used end-to-end with a safe fallback at [queries.go#L1614-L1631](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/queries.go#L1614-L1631), plus the SQL `COALESCE(NULLIF(a.ska_value, ''), v.ska_value)` fix at [addrstmts.go#L242-L246](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/internal/addrstmts.go#L242-L246). The new SKA regression assertion at [balance_test.go#L211-L214](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/balance_test.go#L211-L214) explicitly guards the empty-string path.
- `FromStake > 1.0` is now structurally impossible — the PoS numerator is nested inside the `!isChange` denominator guard at [queries.go#L1551-L1577](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/queries.go#L1551-L1577) (VAR and SKA symmetrically), and the test at [balance_test.go#L191-L195](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/balance_test.go#L191-L195) asserts the bound directly. Stronger than what I suggested.
- `tx_type` namespace comment added at [types.go#L52-L53](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dbtypes/types.go#L52-L53).
- `address.tmpl` re-indented cleanly at [address.tmpl#L126-L154](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/cmd/dcrdata/views/address.tmpl#L126-L154).
- Three new extraction tests cover coinbase reward split, ticket script-fallback classification, and SSFee marker split at [extraction_test.go#L26-L147](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dbtypes/extraction_test.go#L26-L147).

Drive-by `AND valid_mainchain` on `SelectAddressCoinTypes` at [addrstmts.go#L263-L266](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/internal/addrstmts.go#L263-L266) is a good catch — aligns `ActiveCoins` with the balance query and prevents ghost coin types from invalidated chain rows.

### One follow-up, non-blocking

Item #6 (SSGen reward attribution) is still open. `fromStakeVAR` only counts `TxTypeBlockRewardPoS` (101) and `TxTypeSSFeePoS` (103) at [queries.go#L1555](https://github.com/monetarium/monetarium-explorer/blob/47263c5299dd8c767d6dffdbcadd3668b2b97278/db/dcrpg/queries.go#L1555); `stake.TxTypeSSGen` (= 2) is not included. If SSGen outputs carry any PoS payout on Monetarium, stake income is under-counted. A one-liner in the PR description confirming SSFee fully covers PoS payouts (or a follow-up issue) would close this out.

Approving — please address the follow-up in the PR body if it applies before merging.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #321, opened 2026-05-26
.github/workflows: grant id-token write and bump checkout to v6

edshav:
## Summary
- Add `id-token: write` permission so the Claude Code Review job can mint an OIDC token.
- Bump `actions/checkout` from v4 to v6.

## Test plan
- [ ] Trigger the Claude Code Review workflow on a PR targeting `develop` and confirm the job runs to completion without auth or checkout errors.


yanchenko-igor (approved):
LGTM

PR #344, opened 2026-05-27
cmd/dcrdata: tx page revocation Ticket Status; hide empty tspend card

edshav:
## Summary

Fixes #305 — two cosmetic problems on `/tx/{txid}`:

- **Revocation tx was missing the "Ticket Status" indicator** that `/block/{blockhash}` already shows in its Revocations table.
- **Vote tx rendered an empty "Treasury Spend Choices" card** when there were no choices.

Tx page is purely server-rendered (no WS update path), so this is HTML-render only — no JS / pubsub changes.

### Revocation Ticket Status

In `TxPage`, after the existing `if tx.IsTicket() { … }` block, added a sibling branch that calls `exp.dataSource.PoolStatusForTicket(ctx, tx.Vin[0].Txid)` — note the revoked ticket is referenced by Vin[0]'s previous-outpoint hash, not the revocation tx's own hash. Result is stored in `tx.TicketInfo.PoolStatus` (reused because all existing template references to that field are gated by `{{if .IsTicket}}`, so no clash). Error handling is intentionally soft-fail (log + continue) to match the block-page's behavior at `db/dcrpg/pgblockchain.go:6723-6735`.

Template adds a "Ticket Status:" row in the revocation/vote details table, gated by `{{if and .IsRevocation .TicketInfo.PoolStatus}}`.

### Empty Treasury Spend Choices

Moved the `<h5>Treasury Spend Choices</h5>` heading inside the existing `{{if gt (len .TSpends) 0}}` block and dropped the empty-state placeholder. The card now disappears entirely when there are no choices; when there are, it renders unchanged.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...` — clean across all affected modules.
- [x] `golangci-lint run` — 0 issues.
- [x] Pre-commit hooks (gofmt + per-module `go test`) — passed.
- [x] **Playwright manual check on testnet** — visited an `expired` revocation tx (`/tx/0bab9bd2…`); page renders `Ticket Status: expired`, matching the value shown for the same tx in `/block/7055`'s Revocations table.
- [ ] Vote-tx verification (empty + non-empty tspends) — could not exercise on local testnet (no vote txs on chain); template change is mechanically verified.

### DIFF
--- a/cmd/dcrdata/internal/explorer/explorerroutes.go
+++ b/cmd/dcrdata/internal/explorer/explorerroutes.go
@@ -1257,6 +1257,29 @@ func (exp *explorerUI) TxPage(w http.ResponseWriter, r *http.Request) {
 		}
 	} // tx.IsTicket()
 
+	// For a revocation the revoked ticket is referenced by the first input's
+	// previous outpoint, not the revocation tx's own hash.
+	if tx.IsRevocation() && len(tx.Vin) > 0 {
+		ticketHash := tx.Vin[0].Txid
+		zeroHash := (chainhash.Hash{}).String()
+		if ticketHash != "" && ticketHash != zeroHash {
+			_, poolStatus, err := exp.dataSource.PoolStatusForTicket(ctx, ticketHash)
+			if exp.timeoutErrorPage(w, err, "PoolStatusForTicket(revocation)") {
+				return
+			}
+			switch {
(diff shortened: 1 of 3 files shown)


yanchenko-igor (approved):
LGTM

PR #387, opened 2026-06-02
db/cache: normalize genesis block timestamp distorting time-axis charts

edshav:
## Summary
Fixes #385. On networks where block 1 is mined long after block 0 (Monetarium testnet: ~205 days), the genesis timestamp sits months before real network start. Every time-axis chart then begins at the genesis date with all real data compressed into the tail, and the duration-between-blocks chart shows the entire pre-launch gap as its first interval.

The correction lives in the **derived chart cache** (the layer that owns this data) — never in the canonical block timestamp (Postgres `blocks`) or the block-details/API responses, which keep showing the real genesis timestamp.

## Changes
- **`db/cache/charts.go`** — new `NormalizeGenesisBlockTime(times ChartUints) bool` + `genesisGapThreshold` (1h). If `Time[1] - Time[0]` exceeds the threshold it moves `Time[0]` to `Time[1] - 1`. Guards `len < 2` (a genesis-only chain never indexes `Time[1]`); self-no-op after the first correction (resulting gap ~1s < threshold), so incremental syncs and gob reloads are no-ops.
- **`db/dcrpg/queries.go`** — call it in `appendChartBlocks` *after* the height-alignment check, where `Time[0]` is guaranteed to be genesis and only the value (not slice length) is touched.
- **`db/cache/charts.go`** — `cacheVersion` 6.2.0 → 6.3.0 so deployed caches discard the stale `.gob` and rebuild from the DB with the correction applied (no manual cache wipe).
- **`charts_controller.js`** — remove the now-redundant 99th-percentile duration Y-axis clamp; instead anchor the duration Y-floor at 0 with an unbounded top, so genuine large gaps still render at full height and zoomed-in views keep an honest baseline.

One server-side correction fixes the X-axis of every time-axis chart (coin-supply, fees, block-size, chainwork, hashrate, duration) **and** the duration first-interval outlier, since duration is derived in Go from the same `Blocks.Time` array.

## Testing
- New unit tests `TestNormalizeGenesisBlockTime` (long gap, normal gap, threshold boundary, single-block, empty) and `TestNormalizeGenesisBlockTime_Idempotent` — written test-first.
- `db/cache` and `db/dcrpg` (offline) tests pass; both modules `golangci-lint` clean; `gofmt` clean.
- Frontend: prettier clean, eslint clean (only pre-existing warnings), 323/323 vitest pass.
- Verified locally on testnet: full cache recompute from height -1 ran on startup; duration chart first interval is now ~1s, time-axis charts start at real network start, block-0 details still show the real genesis timestamp.

## Notes
- The DB is read-only in this path — no migration or re-sync needed; the `cacheVersion` bump triggers the rebuild automatically.
- Threshold is a documented constant (1h) rather than derived from `TargetTimePerBlock`: `appendChartBlocks` is a free function with no chain params in scope, and `db/cache` does not depend on chainparams.

### DIFF
--- a/cmd/dcrdata/public/js/controllers/charts_controller.js
+++ b/cmd/dcrdata/public/js/controllers/charts_controller.js
@@ -681,11 +681,12 @@ export default class extends Controller {
             false
           )
         )
-        {
-          const vals = data.duration.slice().sort((a, b) => a - b)
-          const maxY = vals[Math.floor(vals.length * 0.99)] || vals[vals.length - 1]
-          gOptions.axes.y = { valueRange: [0, maxY] }
-        }
+        // Anchor the Y-axis floor at 0 (durations are non-negative) but leave the
+        // top unbounded, so genuine large gaps still render at full height and
+        // zoomed-in views keep an honest baseline. The genesis pre-launch gap is
+        // corrected server-side (NormalizeGenesisBlockTime), so the previous
+        // 99th-percentile outlier clamp is no longer needed.
+        gOptions.axes.y = { valueRange: [0, null] }
         break
 
       case 'chainwork': // Total chainwork over time
--- a/db/cache/charts.go
+++ b/db/cache/charts.go
@@ -145,7 +145,7 @@ const (
 // cacheVersion helps detect when the cache data stored has changed its
 // structure or content. A change on the cache version results to recomputing
 // all the charts data a fresh thereby making the cache to hold the latest changes.
-var cacheVersion = semver.NewSemver(6, 2, 0)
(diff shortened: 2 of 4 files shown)


yanchenko-igor (approved):
Solid work. The extraction into a standalone exported function is the right call — it's independently testable, has a clear contract, and the godoc makes it easy to understand why it's safe without digging through the call chain. The inline version works but this is the kind of code you don't second-guess in six months.
Minor cosmetic nit: NormalizeGenesisBlockTime is a mouthful at the call site. Something like NormalizeGenesisGap or NormalizeGenesisTime would read more naturally without losing meaning.

PR #401, opened 2026-06-03
feat: filter out Decred-era agendas (VoteVersion <= 10) from UI and API

yanchenko-igor:
Safety fixes included
- Data race fix (explorerroutes.go:2095): VoteTracker.Summary() returns a pointer to a shared struct — value-copied before mutating Agendas field
- Nil guard (explorerroutes.go:2092-2094): Summary() can return nil before any block is processed; replaced with &agendas.VoteSummary{} to avoid template nil dereference
Tests
- TestAllAgendasVoteVersionFilter (deployments_test.go:300-347): two subtests — one verifying correct threshold filtering and ordering, one verifying empty-slice-without-error when all seeded agendas are below threshold
- TestUpdateAndRetrievals/Test_AgendaInfo (deployments_test.go:264-273): verifies that direct ID lookup still serves pre-Monetarium agendas (filter asymmetry — intentional)
Documentation
- wiki/code-analysis/agendas/compact-flow-and-patterns.md: documents the filter location, ErrNotFound→empty mapping rationale, and filter asymmetry
What this does NOT do
- Does not remove pre-Monetarium agendas from the Storm DB — they remain stored
- Does not block /agenda/{id} or API detail lookups for old agendas (intentional — data still valid, just hidden from lists)
- Does not rename or refactor existing Decred-era naming in the codebase


edshav (approved):
Approving — this correctly closes a gap PR #400 left open: the `/agendas` page renders agendas from **two** sources, and #400 only filtered one of them.

- Table (`agendas.tmpl:246`) ← `AllAgendas()` — filtered in #400.
- Voting-summary cards/meters (`agendas.tmpl:26-219`) ← `voteTracker.Summary().Agendas` — **not** filtered.

This commit cross-filters the cards to the same allowed-ID set so the two surfaces stay consistent.

**Verified:**
- **Key match is valid** — `AgendaSummary.ID` and `AgendaTagged.ID` both derive from `chainjson.Agenda.ID`, so the ID cross-reference is sound.
- **The defensive copy is necessary** — `Summary()` returns the shared, mutex-guarded `tracker.summary` pointer, so `summaryCopy := *rawSummary` before reassigning `.Agendas` avoids mutating shared tracker state. The lock-free read matches the pre-existing pattern (the old code passed that same pointer straight to the template), so no new race.
- **ID cross-reference is the only option** — `AgendaSummary` has no `VoteVersion` field, so filtering against the already-version-filtered list is the right approach.
- Template is safe when `Agendas` ends up empty; builds, `go vet`, `gofmt`, and existing explorer tests all pass.

**When it actually matters:** `summary.Agendas` holds only the highest vote version's agendas (`refreshRCI` keeps the highest successful `GetVoteInfo`). So it's a no-op once the highest version is >= 11, but it removes a real inconsistency while the highest version is still <= 10 (table renders "No agendas found" while the cards still show Decred agendas), and during transient tracker/BoltDB skew.

**Non-blocking suggestions:**
1. **Test coverage** — the filter is currently untested because `voteTracker` is a concrete `*agendas.VoteTracker` (not stubbable in the handler test). Extracting the filter into a small pure helper, e.g. `filterSummaries(summary, allowedIDs)`, would make it unit-testable.
2. A one-line comment on intent would keep a future reader from "simplifying" away what looks like a no-op in steady state.
3. Minor: preallocate `filteredSummaries` with `make([]agendas.AgendaSummary, 0, len(summaryCopy.Agendas))`.

PR #432, opened 2026-06-08
feat: add active miner tracking and rolling-count chart

yanchenko-igor:
This PR introduces end-to-end mining activity tracking for the multi-coin Monetarium network. A new miners table records unique PoW miner addresses and their first/last-seen block heights. The active miner count is displayed on the mining card and as a rolling time-series on the hashrate chart, with an empirical lookback that derives window boundaries from actual block timestamps.
What's included
Database
- New miners table (db/dcrpg/tables.go, db/dcrpg/internal/minerstmts.go) — tracks unique PoW miner addresses with first_seen and last_used block heights.
- Coinbase mining detection (db/dcrpg/pgblockchain.go) — iterates all vouts of all coinbase transactions and classifies PoW-reward-type outputs (P2PKH, P2SH, P2PK, alt-pubkey) as miner addresses.
- SelectHeightByTimestamp query (db/dcrpg/internal/blockstmts.go) — exact height lookup for any Unix timestamp.
- Backfill migration — schema upgrade threads through fresh installs and existing deployments at any prior maintenance level.
Backend (db/cache, db/dcrpg)
- MinerRange data type and SetMinerRanges method on ChartData — stores miner first-seen/last-used in the charts cache.
- activeMinersCounts — event-sweep algorithm computes per-block active miner count using empirical block timestamps (binary search on charts.Blocks.Time) instead of hardcoded block counts. Supports Day, Week, Year, and All intervals.
- Hashrate chart integration — chart endpoint returns active_miners series alongside hashrate.
- Context propagation — upsertMiner now uses ExecContext with proper context.Context propagation for graceful shutdown.
- Home page active miner count — uses GetHeightByTimestamp(now - 7d) for the week-lookback threshold, consistent with the chart's empirical approach.
Frontend
- Hashrate chart dual-axis — Y1: Network Hashrate, Y2: Active Miners (left/right axes).
- Clean integer ticker for the Y2 axis — no fractional labels, no duplicate tick marks.
- Interval selector — Day/Week/Year/All toggle for the active miner lookback window.
- Series visibility toggles — show/hide hashrate or active miners independently.
- Dual-series legend formatter — separate formatting for hashrate (Ph/s) and miner count (integer).
- Enabled chart caching — active miner chart data is cached alongside other chart series.
- Mining card — shows active miner count with correct singular/plural ("1 miner" vs "N miners").
UI
- home_mining.tmpl — simple pluralization fix.
Verification
- All 327 frontend tests pass (vitest, jsdom).
- Linter: 0 errors (npm run check).
- Go build succeeds across all 3 affected modules.
- Go vet clean.
- db/cache package tests pass.
- Schema upgrade tested for both fresh installs and existing deployments.


edshav (commented):
## Code review — active-miner tracking & rolling-count chart (#424 / #425)

Reviewed 22 commits vs the `develop` merge-base (18 files). All modules compile; call sites, `explorerDataSource` implementers, and the `home_mining.tmpl` dot-context (`*types.HomeInfo`) all check out. Findings are ranked by confidence × severity.

Context worth stating up front: the `miners` table stores only `(first_seen, last_used)` per address — a **single interval** — but #424 asks for a **rolling-window unique-address count**. You can't reconstruct "did address X mine within [B−week, B]" from a min/max pair, so several findings below are consequences of that lossy data model rather than isolated typos.

> _Edit: an earlier version of this review listed a third ship-blocker — `BackfillMiners`' `value > 0` filter "dropping SKA-only miners." That's retracted: SKA coins aren't mineable, so every coinbase is a VAR (single-coin) tx and all miner-reward outputs have `value > 0`. The filter excludes nothing real and is correct._

---

### 🔴 High-confidence correctness bugs

**1. `activeFull[HashrateAvgLength:]` panics on chains with 2–119 blocks** — `db/cache/charts.go:1681`
The BlockBin guard only rejects `len(Blocks.Time) < 2`, but `hashrate()` needs ≥120 blocks and `activeMinersCounts` returns a slice of length `numBlocks`. For any `2 ≤ numBlocks < 120` (fresh testnet, local short-chain dev DB, mainnet mid-sync), `activeFull[120:]` → `slice bounds out of range [120:N]`, which 500s `GET /api/chart/hashrate`. The old code never indexed past the rate length, so this is newly introduced.
*Fix:* only slice when `numBlocks > HashrateAvgLength`, else use an empty slice — matching `hashrate()`'s own empty-return path.

**2. The `miners` table is not reorg-aware** — `db/dcrpg/pgblockchain.go:3955` (live insert) + `TipToSideChain` at `:3686`
Two facets, one root cause:
- The coinbase upsert loop runs unconditionally inside `StoreBlock`, **not gated on `isMainchain`**, so side-chain blocks (`cmd/dcrdata/main.go:911`, the side-chain import path) record their coinbase address as a miner. `BackfillMiners` filters `is_mainchain = true`, so the live path and the backfill disagree.
- When a mainchain block is later orphaned, `TipToSideChain` reverts blocks/txns/vouts/votes but **never rolls back `miners`** — `last_used`/`blocks_mined` stay at the now-orphaned height. `UpsertMinerRow` also sets `last_used = EXCLUDED.last_used` unconditionally, so a reorg to a lower height can move `last_used` backward.

Net: after any reorg the table permanently over-counts and diverges from what a backfill would produce, corrupting both the home threshold and the chart.
*Fix:* gate the loop on `isMainchain`; revert/recompute affected miner rows in `TipToSideChain`.

---

### 🟠 Spec-conformance (medium)

**3. The rolling window is approximated as one continuous interval — over-counts mid-chart** — `db/cache/charts.go` (`activeMinersCounts`)
The sweep emits `+1` at `FirstSeen` and `−1` only at `time[LastUsed] + window`, so a miner is counted active for **every** block between its first and last appearance, gaps included. A miner that mined at height 100 and again 6 months later is counted active for the entire idle gap. #424 counts a miner at block B only if it mined within the lookback ending at B. Confirmed deviation; mid-history values are inflated by the number of idle/address-changing miners — the exact double-counting the Interval toggle was meant to prevent.

**4. Home value won't reliably equal the chart's Week tip — violates #425 "must equal"** — `cmd/dcrdata/internal/explorer/explorer.go:~537`
Home counts `last_used >= heightOf(tip−7d)` (height axis, inclusive `>=`); the chart expires a miner when `time[h] > time[LastUsed] + 7d` (time axis, strict `>`). When `time[minHeight] < tip−7d` (the common case), a miner with `last_used == minHeight` is counted by home but already expired on the chart — an off-by-one band. The deeper (altitude) issue: **two independent definitions** of the same quantity (`activeMinersCounts` event-sweep vs `CountActiveMiners` SQL) are maintained separately and can only drift. Consider one shared definition both surfaces consume.

**5. Independent updaters can desync miner heights from `Blocks.Time`** — `db/cache/charts.go:947`
`minerRanges` and `chartBlocks` are separate `ChartUpdater`s. `activeMinersCounts` bounds-checks `LastUsed` but **not `FirstSeen`**, and a miner whose `LastUsed` exceeds the loaded block range gets no exit event (counted active forever). If the miners table is ahead of the loaded `Blocks.Time` slice, the sweep is unbalanced — over- or under-counting.
*Fix:* clamp/guard both endpoints symmetrically.

---

### 🟡 Robustness / silent failures (medium)

**6. `ActiveMiners` DB error renders "Mining: 0 miners"** — `cmd/dcrdata/internal/explorer/explorer.go:~543`
On error the value is only logged; `activeMiners` keeps its zero value and unconditionally overwrites `p.HomeInfo.ActiveMiners`. A transient timeout shows `0` instead of the last-known count. Prefer leaving the prior value untouched on error.

**7. `GetHeightByTimestamp` ErrNoRows on chains younger than the lookback** — `cmd/dcrdata/internal/explorer/explorer.go:~537`
When `tip−7d` predates genesis, `SELECT … LIMIT 1` returns no rows → `sql.ErrNoRows` → `minHeight = 0` → counts *all* miners, plus a warning logged on **every** new block. Harmless on a mature mainnet, but noisy and wrong for the first week / short-chain testnets. Treat ErrNoRows as "height 0, no warning."

**8. Live miner count is stale until a full page reload** — `pubsub/pubsubhub.go` (`GeneralInfo`) + `cmd/dcrdata/public/js/controllers/mining_controller.js`
#425 asks to "wire the live-update path." Two gaps: the pubsub `GeneralInfo` payload never sets `ActiveMiners`, and the WS-driven `mining_controller` has no target for the `Mining: {N} miners` heading (server-rendered only). Sibling metrics (hashrate, difficulty) update live while the miner count freezes. Either add a `data-mining-target` + populate the WS field, or accept reload-only and document it.

---

### 🔵 Lower-severity / operational

**9.** `BackfillMiners` runs under a hard 30-min timeout during the schema 0→1 upgrade (`db/dcrpg/upgrades.go:263`) — on a large mainnet `vouts⋈transactions` GROUP BY this can `DeadlineExceeded` and block startup. Idempotent on retry, but consider a config knob or a background backfill.

**10.** DayBin `dailyActive[i]` stays `0` when the day's height ≥ `len(activeFull)` (`db/cache/charts.go:1706`) — a visible dip at the right edge of the day-binned chart.

---

### ⚪ Optional cleanup (no action required)

- **Upgrade latent risk:** schema 0→1 sets `schema=1` but never resets `maint` before falling through to `case 1`; works today only because the new maint numbering coincides with the old. The deleted `default: unsupported schema version` branch now silently passes unknown schemas through `upgradeCheck()`. (`db/dcrpg/upgrades.go`)
- `rawDataURLTarget` now appends `&interval=week` to **every** chart's URL, not just hashrate (`charts_controller.js:824`) — misleading but inert (backend forces `DefaultInterval`).
- Hashrate render rebuilds `this.visibility` from checkbox defaults, ignoring a `?visibility=` value restored from the URL (`charts_controller.js:777`) — bookmarked series-visibility state is lost.
- Dead/limited code: `CountMiners` (`queries.go:5101`) has no callers; the `default:` arm in `activeMinersCounts` duplicates `WeekInterval`; `SetMinerRanges` is a one-line setter whose "hold write lock" contract the caller doesn't actually route through.
- Efficiency: during initial sync, per-block `upsertMiner` round-trips duplicate the bulk `BackfillMiners`; `Store()` adds two serial queries per block on the hot path — both cacheable.

---

**Bottom line:** #1 and #2 are the ship-blockers (a crash and reorg corruption). #3–#5 are real but stem from the single-interval data model — worth deciding whether the approximation is acceptable for the client or whether per-block recipient data is needed. The rest is robustness polish.

edshav (commented):
### Re-review — against review #4460321120

5 fix commits landed after the prior review (`7b121fbc0` → `1291672e6`). Most findings are resolved; the live-update (#8) and reorg (#2b) fixes each introduced a **new backend issue**, and spec items #3–#5 are largely unchanged because they stem from the single-interval `(first_seen, last_used)` data model.

## New issues introduced by the fix commits

**A. `pubsubhub.Store` runs two DB queries while holding the pub/sub write lock** (contention regression)

The live-update fix added `GetHeightByTimestamp` + `ActiveMiners` calls **inside** `p.mtx.Lock()` (taken at `pubsubhub.go:700`, released at `:968`). Every WS subscriber reading `p.GeneralInfo` is blocked for both Postgres round-trips on every new block. The sibling fix in `explorer.go` does it correctly — queries run *before* `p.Lock()`. Fix: hoist both calls above the lock.

https://github.com/monetarium/monetarium-explorer/blob/1291672e62b9fa10adb343f7bac2a9fa74c318d4/pubsub/pubsubhub.go#L725-L734

**B. `RevertOrphanMinerUpdate` matches by `block_height` only — over-decrements on repeated same-height reorgs**

The revert SQL selects coinbase addresses with `WHERE t.block_height = $1` and **no `block_hash` filter**. In `TipToSideChain` it runs *after* steps 1–2 set `is_mainchain=false`, so neither the height nor the mainchain flag distinguishes the just-orphaned block from other blocks stored at that height (orphaned blocks remain in `transactions`). A second reorg at a height that previously saw one can decrement `blocks_mined` for a tracked miner whose coinbase sits in an unrelated stored block at that height — and `CleanupMinerZeros` then deletes it. Pass `tipHash` and add `AND t.block_hash = $2`.

https://github.com/monetarium/monetarium-explorer/blob/1291672e62b9fa10adb343f7bac2a9fa74c318d4/db/dcrpg/internal/minerstmts.go#L58-L69

https://github.com/monetarium/monetarium-explorer/blob/1291672e62b9fa10adb343f7bac2a9fa74c318d4/db/dcrpg/pgblockchain.go#L3775-L3787

## Prior findings — current status

| # | Finding | Status |
|---|---------|--------|
| 1 | Short-chain slice panic | ✅ Fixed — `if numBlocks > HashrateAvgLength` guard (`charts.go:1685`) |
| 2a | Side-chain coinbases recorded as miners | ✅ Fixed — upsert gated on `isMainchain` |
| 2b | Reorg `last_used` rollback | 🟠 Partial — `blocks_mined` reverted; `last_used` is a documented approximation (`$1-1`, not true prior height). See **B**. |
| 3 | Rolling window over-counts idle gaps | 🔴 Open by design — inherent to the `(first_seen, last_used)` schema; not documented as an accepted approximation |
| 4 | Home value ≠ chart Week tip (#425 "must equal") | 🔴 Open — still two definitions (`CountActiveMiners` SQL `>=` height-axis vs sweep `>` time-axis); fixes only touched error handling |
| 5 | Endpoint bounds asymmetry | 🟠 Partial — `FirstSeen` now guarded (`charts.go:949`), but `LastUsed >= numBlocks` still drops the exit event → counted active forever in the desync case |
| 6 | DB error renders "0 miners" | ✅ Fixed — assignment gated on `err == nil` (minor: warning log dropped, failures now silent) |
| 7 | `ErrNoRows` on young chains | ✅ Fixed — wrapper returns `(0, nil)` |
| 8 | Live count stale until reload | ✅ Fixed — wired end-to-end; `0` and pluralization OK (but see **A**) |
| 9 | Backfill timeout | 🟢 Bumped 30 min → 1 h |
| 10 | DayBin right-edge dip | ⚪ Cosmetic, unchanged |

**Minor / optional:** DayBin time-axis hashrate response omits the `h` height field CLAUDE.md asks for on time-axis chart endpoints (`charts.go:1713-1723`); the dropped warning log in #6; PR title uses `feat:` rather than the `package/path:` convention. None are blockers.

**Bottom line:** the two ship-blockers are addressed (#1 fully, #2 functionally with caveats). Genuinely actionable now: the two new backend issues — **A** (clear, easy fix) and **B** (narrow but real) — plus still-open **#4** if #425's "must equal" is a hard requirement. **#3/#5** remain consequences of the single-interval schema; worth an explicit "accepted approximation" decision + doc comment rather than code changes.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

edshav (commented):
### Re-review — against review #4461033882

3 new commits (`d3bdbfef6`, `bf2810a48`, `9b34d4e0e`). All five items from the prior review are resolved; one new WS-path regression was introduced by the lock-hoist.

## Prior items — all fixed

| Item | Status |
|---|---|
| **A** — pubsub DB queries held the write lock | ✅ Fixed — both queries now run *before* `p.mtx.Lock()` (`pubsubhub.go:697-716`) |
| **B** — `RevertOrphanMinerUpdate` matched by `block_height` only | ✅ Fixed — SQL now `WHERE t.block_height = $1 AND t.block_hash = $2`; call site passes `orphanHeight, tipHash` (a `ChainHash`, BYTEA-correct); CASE arm still compares `last_used = $1` (height). Param order correct. |
| **#4** — home count vs chart Week-tip | ✅ Fixed — `CountActiveMiners` flipped `>= $1` → `> $1`, removing the off-by-one; the residual height-vs-time axis discrepancy is acknowledged as inherent in a new comment |
| **#5** — `LastUsed` not clamped (dropped exit event) | ✅ Fixed — clamps `used = numBlocks-1`, always emits exit event; index-safe |
| **#3** — single-interval over-count | ✅ Documented — accurate comment at `charts.go:925-930` |

## New regression introduced by the lock-hoist (A's fix)

`pubsubhub.Store` now clobbers the last-known active-miner count with `0` on a transient DB error. `activeMinersCount` is zero-initialized and only set on success, but the assignment is unconditional:

```go
var activeMinersCount int64                    // 693: zero
if count, err := psh.sourceBase.ActiveMiners(ctx, minHeight); err == nil {
    activeMinersCount = count                  // only on success
}
...
p.GeneralInfo.ActiveMiners = activeMinersCount // 738: always writes — 0 on error
```

This is the same bug class as the original finding #6 — and the HTTP path still guards it correctly (`explorer.go:571`, `if err == nil`), so on a transient error the two paths now disagree and the WS push flashes "Mining: 0 miners". Hoisting the query out of the lock dropped the success guard that used to wrap the assignment. Fix: carry an `ok` bool and gate line 738 on it (the prior value can't be read pre-lock without a race, so a sentinel is the race-free way to mean "leave unchanged").

https://github.com/monetarium/monetarium-explorer/blob/9b34d4e0edcb8e06521a6ac72567eb1e75790e29/pubsub/pubsubhub.go#L691-L705

https://github.com/monetarium/monetarium-explorer/blob/9b34d4e0edcb8e06521a6ac72567eb1e75790e29/pubsub/pubsubhub.go#L736-L738

## Minor

The `CountActiveMiners` doc comment still says "at or above" — stale after the `>=`→`>` change.

https://github.com/monetarium/monetarium-explorer/blob/9b34d4e0edcb8e06521a6ac72567eb1e75790e29/db/dcrpg/queries.go#L5106-L5110

**Bottom line:** clean iteration — every flagged item fixed. Only the WS-path zero-clobber regression is worth addressing before merge; the stale comment is a one-liner.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

edshav (approved):
### Re-review — against review #4461033882

One new commit since the prior re-review (#4461414784, @ `9b34d4e0`): `b1ec707e5` "guard pubsub ActiveMiners write with sentinel, fix stale doc comment". Both findings from #4461414784 are resolved; one sibling stale comment was missed (cosmetic). Approving.

## Prior findings — status

| From | Finding | Status |
|---|---------|--------|
| #4461414784 | WS-path zero-clobber (flashes "0 miners" on transient DB error) | Fixed — `activeMinersOK` is set only on success and now gates the write. `p := psh.state` is persistent hub state, so on a transient `ActiveMiners` error the prior value is left untouched (race-free; no shared read before the lock). |
| #4461414784 | Stale `CountActiveMiners` doc comment ("at or above") | Fixed — now "strictly above", matching `last_used > $1`. |
| #3 / #5 | Single-interval `(first_seen, last_used)` approximation | Accepted/documented, unchanged. |

Sentinel gate and the fixed comment:

https://github.com/monetarium/monetarium-explorer/blob/b1ec707e5f494e6b7c1568bfc13f6217ab9c7dd4/pubsub/pubsubhub.go#L739-L742

https://github.com/monetarium/monetarium-explorer/blob/b1ec707e5f494e6b7c1568bfc13f6217ab9c7dd4/db/dcrpg/queries.go#L5107-L5110

## Minor (non-blocking)

The sibling doc comment on the `(*ChainDB).ActiveMiners` wrapper still reads "at or above the given minimum block height", while the SQL it delegates to is `WHERE last_used > $1` (strictly above). Commit `b1ec707e5` — titled "fix stale doc comment" — fixed the identical comment on `CountActiveMiners` one layer down but left this one. One-line cosmetic follow-up; no runtime impact.

https://github.com/monetarium/monetarium-explorer/blob/b1ec707e5f494e6b7c1568bfc13f6217ab9c7dd4/db/dcrpg/pgblockchain.go#L5392-L5396

Already raised in earlier reviews and not re-flagged: PR `feat:` title vs the `package/path:` convention (#4461033882); the missing `h` height field on the time-axis hashrate response (#4461414784, "Minor/optional").

**Bottom line:** clean iteration — both #4461414784 findings are correctly fixed, the sentinel is race-safe, and no new bugs were introduced. The only residual is the cosmetic sibling-comment miss above. Approving.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #456, opened 2026-06-11
blockdata: fix MiningFee to use miner VAR fee, not total per-tx sum

yanchenko-igor:
## Summary

`computeMiningFee` was summing fees across all regular + stake transactions, reporting total VAR fees (52,270 atoms for block 4423) as `MiningFeeAtoms`. This is wrong — the miner only collects 50% of total VAR fees (26,135 atoms), with the other 50% going to the ticket seller. The old code attributed 100% of fees to the miner, including the ticket seller's 50% share of ticket purchase fees.

Replaced with `computeMinerVARFeeAtoms`, which reads the miner's fee from the coinbase by finding the highest-value P2PKH/P2SH output and subtracting the PoW subsidy (giving 26,135 = 50% of 52,270). Uses max-value selection to avoid false positives from non-miner outputs (e.g., commitment, OP_RETURN, legacy P2SH) that may appear before the miner output.

## Changes

| File | Change |
|------|--------|
| `blockdata/blockdata.go` | Replaced `computeMiningFee` with `computeMinerVARFeeAtoms`; max-value output selection; updated caller in `CollectBlockInfo` |
| `blockdata/blockdata_test.go` | Renamed all tests; added `CoinbaseNoP2PKH` edge case; added synthetic 3-output coinbase test mimicking block 4423 (nonstandard, OP_RETURN, P2PKH); fixed misleading comment about ticket fees |
| `cmd/dcrdata/internal/explorer/explorer_test.go` | Removed JSON fixture loading; uses synthetic block |

## Test Plan

- [x] `go test ./blockdata/` — 8 synthetic tests validate fee logic, 1 synthetic 3-output test validates skip logic
- [x] `go test ./cmd/dcrdata/internal/explorer/` — existing `TestStore_MiningFeeFromRealBlock4423` verifies `MiningFeeAtoms = 26_135` through the Store pipeline
- [x] `go test ./...` across all 3 modules — all green

Root cause: `computeMiningFee` summed ALL fees (100%) as if the miner gets everything. In Monetarium, total VAR fees (regular + ticket purchase) are split 50/50 between miner and ticket seller — the miner's share is `coinbase_P2PKH_value - PoW_subsidy = 26,135` atoms (50% of 52,270). The fix reads the coinbase output directly rather than summing per-transaction fees.

### DIFF
--- a/api/types/apitypes.go
+++ b/api/types/apitypes.go
@@ -809,8 +809,9 @@ type BlockExplorerExtraInfo struct {
 	CurrentBlockSubsidy *chainjson.GetBlockSubsidyResult `json:"current_block_subsidy"`
 	NextBlockSubsidy    *chainjson.GetBlockSubsidyResult `json:"next_block_subsidy"`
 	// MiningFeeAtoms is the miner's 50% share of VAR transaction fees (not total
-	// block fees). Read from the coinbase P2PKH/P2SH output minus the vote-scaled
-	// PoW subsidy. The other 50% goes to the ticket seller via SSFee.
+	// block fees). Computed as Σ(VAR coinbase outputs) − Σ(coinbase inputs),
+	// which matches the tx page's FeeReward by construction. The other 50%
+	// goes to the ticket seller via SSFee.
 	MiningFeeAtoms int64 `json:"mining_fee_atoms"`
 	// CoinAmounts holds per-coin totals (VAR key=0, SKAn key=n) as decimal atom strings.
 	CoinAmounts map[uint8]string `json:"coin_amounts,omitempty"`
--- a/api/types/apitypes.go
+++ b/api/types/apitypes.go
@@ -808,7 +808,10 @@ type BlockExplorerExtraInfo struct {
 	CoinSupply          int64                            `json:"coin_supply"`
 	CurrentBlockSubsidy *chainjson.GetBlockSubsidyResult `json:"current_block_subsidy"`
 	NextBlockSubsidy    *chainjson.GetBlockSubsidyResult `json:"next_block_subsidy"`
-	MiningFeeAtoms      int64                            `json:"mining_fee_atoms"`
+	// MiningFeeAtoms is the miner's 50% share of VAR transaction fees (not total
+	// block fees). Read from the coinbase P2PKH/P2SH output minus the vote-scaled
+	// PoW subsidy. The other 50% goes to the ticket seller via SSFee.
+	MiningFeeAtoms int64 `json:"mining_fee_atoms"`
 	// CoinAmounts holds per-coin totals (VAR key=0, SKAn key=n) as decimal atom strings.
 	CoinAmounts map[uint8]string `json:"coin_amounts,omitempty"`
 	// CoinTxStats holds per-coin tx count and size (key 0=VAR, 1-255=SKAn).
--- a/blockdata/blockdata.go
+++ b/blockdata/blockdata.go
@@ -17,7 +17,6 @@ import (
 	"github.com/monetarium/monetarium-node/cointype"
 	"github.com/monetarium/monetarium-node/dcrutil"
 	chainjson "github.com/monetarium/monetarium-node/rpc/jsonrpc/types"
-	"github.com/monetarium/monetarium-node/txscript/stdscript"
 	"github.com/monetarium/monetarium-node/wire"
 
 	apitypes "github.com/monetarium/monetarium-explorer/api/types"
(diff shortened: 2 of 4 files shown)


edshav (commented):
Reviewed the approach end-to-end. The core instinct is right — the miner's fee lives in the coinbase, and reading it from there is the correct direction — and the PR correctly leaves the Block# "Fees" header untouched. But I found one correctness bug and one test gap that I think should be addressed before merge.

## 🔴 Wrong fee (shows `0`) for blocks with fewer than 5 votes

`CollectBlockInfo` calls `computeMinerVARFeeAtoms(msgBlock, curSubsidy.PoW)` (`blockdata/blockdata.go:251`), where `curSubsidy` is `GetBlockSubsidy(ctx, height, 5)` (`blockdata/blockdata.go:201`) — the voter count is **hardcoded to 5**. The function returns `minerOutputValue - powSubsidy`, clamped to `0` when `minerOutputValue <= powSubsidy`.

The issue: the **work subsidy scales with the number of votes**. It's computed via `CalcWorkSubsidyV3(height, votesPerBlock, …)` (`txhelpers/subsidy.go:63`), and the comment on the next line confirms `votesPerBlock` (5) is the *maximum*. So `curSubsidy.PoW` is always the **full** subsidy, whereas the actual coinbase output carries the **vote-scaled** subsidy (`subsidy × voters / 5`) for blocks above `StakeValidationHeight` (4096).

Worked example — a post-4096 block with 4 votes:

- actual coinbase miner output = `fullSubsidy × 4/5 + fee`
- `powSubsidy` passed in = `fullSubsidy`
- `minerOutputValue − powSubsidy` = `fee − fullSubsidy/5` → large negative → clamped to **`0`**

`fullSubsidy/5` is on the order of several VAR while the fee is a fraction of a VAR, so the home page would display **Fee: 0** for that block. Blocks with 3–4 votes are routine on mainnet, so this is "the most-recent block had a missed vote → Fee reads 0," not a rare corner case.

**Suggested fix:** subtract the coinbase's *actual* input value rather than the hardcoded-5-vote RPC subsidy — or equivalently compute `Σ(coinbase outputs) − Σ(coinbase inputs)`. The coinbase input already carries the real vote-scaled subsidy, so that form is correct for any vote count and also matches the PoW-Reward tx page's `FeeReward` (`outputs − inputs`) by construction.

## 🔴 The headline test doesn't exercise the new function

`TestStore_MiningFeeFromRealBlock4423` passes `msgBlock := &wire.MsgBlock{}` (empty, no coinbase) and pre-sets `ExtraInfo.MiningFeeAtoms: 26_135`. `Store` never calls `computeMinerVARFeeAtoms` (that happens in the *collector*), so this test only proves `Store` copies a value it was handed — there's no real block, no coinbase, and no subsidy subtraction under test, despite the name. The `blockdata_test.go` synthetic tests do call the function, but they choose the output value and the subsidy to be mutually consistent, so they can't catch the 5-vote mismatch either. A test that feeds a vote-scaled coinbase output against a `GetBlockSubsidy(…, 5)` subsidy (e.g. a real 3- or 4-vote block) is what's missing.

## Against the issue's acceptance criteria

| # | Criterion | Verdict |
|---|-----------|---------|
| 3 | Block# "Fees" header unchanged | ✅ Confirmed — that header is computed independently in `db/dcrpg/pgblockchain.go:6868` (`getTotalFee(Tx)+getTotalFee(Tickets)`) and does not read `ExtraInfo.MiningFeeAtoms`. |
| 1 | Equals the PoW-Reward tx page Fee Reward (`outputs − inputs`) | ⚠️ Diverges — this computes `maxOutput − fullSubsidy`, which agrees only for single-output, exactly-5-vote coinbases; it returns `0` in the <5-vote case above. |
| 2 | #19103 reads `0.0001359` | ✅ Likely — *if that block had 5 votes* — but not guaranteed by the tests. |

## Minor / non-blocking

- **Max-output heuristic** (`computeMinerVARFeeAtoms`): selecting the largest P2PKH/P2SH VAR output assumes exactly one miner output that is always the largest and an unsplit subsidy. Fine for today's coinbase shape, but it's more assumptions than a conservation-based `outputs − inputs`.
- **Repurposing `ExtraInfo.MiningFeeAtoms`** from "full block fees" to "miner's half" is contained — the only readers are the home-page and websocket paths and it isn't served by any `/api` endpoint — but the field's meaning changed while its name and doc comment did not. Worth a comment update.

## Net

The direction is sound and the Block# header handling is correct. The two blocking items are the hardcoded-5-vote subsidy (which regresses the very metric being fixed for a common class of blocks) and a headline test that bypasses the new code. Addressing those — ideally by subtracting the actual coinbase input / using `outputs − inputs`, plus a genuine <5-vote test — would get this to approval.

edshav (approved):

edshav (approved):

